PR #5, opened 2016-01-28
Feature/binary search

aziz-boudi4:
Here is an implementation of Binary Search that I made while I was learning algorithm


hollance:
Thanks for contributing this. I was already in the process of writing an explanation for binary search and my code is slightly different from yours. I did use your array of prime numbers, though. :-)

PR #7, opened 2016-01-28
Updates pop() to use `guard`

micpringle:


hollance:
Thanks for the PR, Mic, but I don't really see the benefit of using `guard` here. IMHO, it doesn't really make things clearer. I'm inclined to stick to the good old `if` here.

Unless you can convince me otherwise, of course. :-)

darren102:
Would the consideration instead of using count to use the isEmpty method instead for checking?

micpringle:
I think, in this case, it's more a preference of style than anything else. That, and using the latest language features. So, I'm afraid I don't have a great argument to persuade you with. :]

hollance:
@darren102 You mean `if isEmpty { ... }`? That sounds reasonable. ;-)

darren102:
@hollance thinking `if array.isEmpty { ... }` for the check since not sure of underlying array implementation but if it were implemented badly and the count not stored then it would be dependent upon the size of the array where `isEmpty` just needs to verify if there are elements or not.

PR #8, opened 2016-01-28
Using isEmpty instead of count for the array pop method

darren102:
Per discussion previously using `isEmpty` on the array instead of `count`

### DIFF
--- a/Stack/Stack.swift
+++ b/Stack/Stack.swift
@@ -19,11 +19,9 @@ public struct Stack<T> {
   }
 
   public mutating func pop() -> T? {
-    if count == 0 {
-      return nil
-    } else {
-      return array.removeLast()
-    }
+    guard !array.isEmpty else { return nil }
+
+    return array.removeLast()
   }
 
   public func peek() -> T? {


kevinrandrup:
That guard is really confusing. It reads like "if array is not 'not empty' then return nil. `isEmtpy` is better than `count == 1` but I think guard should be reserved for failing to bind an optional value at the beginning of the function.

hollance:
I like the use of `isEmpty`, but agree that `guard` makes it more confusing than it needs to be. So I'll close this PR and make those changes.

micpringle:
Why not simply...

```swift
if array.isEmpty { return nil }
return array.removeLast()
```

I don't think an `else` is necessary here.

hollance:
The `else` is not strictly necessary, but in my opinion it does clarify the flow of what's going on.

micpringle:
In that case, what about ...

```swift
return array.isEmpty ? nil : array.removeLast()
```

:smile:

hollance:
Perfectly fine, but since the goal of this project is to teach algorithms I'd prefer to stick with an simple `if-else` since everyone understands that.

PR #9, opened 2016-01-29
Conforms to ArrayLiteralConvertible

samirGuerdah:
Add the ArrayLiteralConvertible protocol conformance.

### DIFF
--- a/Stack/Stack.playground/Contents.swift
+++ b/Stack/Stack.playground/Contents.swift
@@ -11,7 +11,7 @@
   Push and pop are O(1) operations.
 */
 
-public struct Stack<T> {
+public struct Stack<T> : ArrayLiteralConvertible {
   private var array = [T]()
   
   public var count: Int {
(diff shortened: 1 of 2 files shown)


hollance:
Thanks, Samir. However, I'm a bit torn about this. It's great to have `ArrayLiteralConvertible` for things that can be initialized with arrays but usually you start out with an empty stack or queue, so I'm not sure it's worth doing here.

If we were to do it, I'd like to put it in an extension so it doesn't clutter up the main struct. However, that seems to crash my version of the Swift compiler...

samirGuerdah:
Hello,

Ok, i think that you are right to keep it simple and focus in the algorithmic part. 
Just one note, you can still create an empty Stack even if you conform to the ArrayLiteralConvertible
```
let emptyStack = Stack<Int>()
```
The playground crashes also if i do it in an extension :). I think that we can close it.

hollance:
All right. Thanks for the contribution, though! :-)

PR #14, opened 2016-02-02
Two-Sum O(n) solution without needing to previously sort array

D-Speiser:


hollance:
Hey, thanks for that! I have a request to make, though: instead of replacing the existing solution, could you simply add this new one as an alternative? It's nicer for readers to see two solutions.

Can you also add a description of your solution to the README, which an explanation of the upsides and downsides compared to the other algorithm? That would be awesome, thanks!

D-Speiser:
Sure, not a problem. I'll push changes later tonight or tomorrow.

D-Speiser:
pushed changes you've asked for @hollance

hollance:
Just a quick update: I edited your description slightly to fit in better with the existing text.

I also made a small change to the code. Instead of `for i in 0 ... a.count - 1`, I used `for i in 0 ..< a.count`. It's debatable which one is "better", but I prefer the latter.

Thanks again for this!

PR #15, opened 2016-02-02
Added a basic implementation of an unbalanced btree using enums

nameghino:
Just what the title says, the basic implementation for a non-balancing btree using generic enums.
The generic type is restricted to the Comparable protocol to allow comparison for insertion.

Let me know any comments you have on this!


hollance:
Nice! Could you write up an explanation too?

Note: I would call it "binary tree" instead of btree because a B-tree is a different kind of data structure. 

Technically, this is even a "binary search tree" because you're ordering the nodes. So I'd prefer it if the folder was called "Binary Search Tree" rather than just "Trees".

nameghino:
Sure, should I write the explanation in the comments on the file or in the readme?

digitsu1:
I would recommend both that way even if reader has only file can understand.

hollance:
The explanation should go into README.markdown. Be sure to also add your name to the file, so you get credit for your work.

nameghino:
Sorry I had to make a lot of commits, kept forgetting stuff. The explanation lives on a README file inside the BST directory, like the Stack has.
As a TBD, I'll add a `delete` method and its explanation.

Let me know your thoughts!

donald-pinckney:
@hollance This isn't really related to this PR, but on the subject of actual B or B+ trees, is that something you might like added?  Though they are for pointing to disk blocks on hard drives and such, you can still implement them all in memory as a tool for teaching how they work.

hollance:
@nameghino Nice! Thanks for this. I'm looking forward to the `delete` method (it's the tricky one!). :-)

@donald-pinckney I'd absolutely like to have a B / B+ tree. even if it's just for teaching.

hollance:
@nameghino I've tweak the text a little (nothing major). So if you're still working on this, you may want to merge those changes first.

hollance:
@nameghino FYI, I've added a class-based binary search tree and a whole bunch of new explanations. 

If you want, you can still add code to delete nodes from your enum-based tree.

nameghino:
Thanks, I'll merge!

PR #16, opened 2016-02-03
Binary Tree

mitghi:
I added Binary Tree implementation. It uses Generics and can be used right-away as key-value data structure.

Example:

    //                   key, value ( payload )

    var mbt = BinaryTree<Int,String>()

    for i in [10,8,9,24,22,26]{
        mbt[i] = "\(i)_test"
    }

    mbt.delete(10)
    print("\(mbt.root!.rightchild!.rightchild!.key) -> \(mbt.root!.rightchild!.rightchild!.value)")
    mbt.delete(26)
    print(mbt.root!.rightchild!.rightchild?.value)

### DIFF
--- a/Binary-Tree/binary_tree.swift
+++ b/Binary-Tree/binary_tree.swift
@@ -1,0 +1,266 @@
+// The MIT License (MIT)
+
+// Copyright (c) 2016 Mike Taghavi (mitghi[at]me.com)
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+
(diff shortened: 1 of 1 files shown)


hollance:
Thanks! Someone else is already providing a binary search tree but their implementation uses an enum and doesn't have all your functionality.

Once their version is added to the repo, could you add your code in that same folder? And also write an explanation of how your version works? That way readers can compare the different approaches, that would be great.

mitghi:
Sure, I will do that. Should i close this pull request ?

hollance:
Yeah, I'll close it.

mitghi:
Please provide some additional information about the folder structure ? How should i put it there ? Should i add another readme ?

hollance:
Ideally, I'd like to have all the explanations in a single README file. The existing version uses an enum. So you'd add a new section to the README explaining how your version is different, and explain any additional functionality that you added.

There is a small problem: the other binary search tree README isn't finished yet. So it's probably a bad idea to start messing with those files right now.

Here's an idea: Create a folder "Binary Search Tree 2" (in the top-level directory) and add your stuff in there. Then at some point I will merge the READMEs.

How does that sound?

mitghi:
Thanks a lot for the info. That sounds great, but i just rewrote it as an AVL tree. You can have look at it (here)[https://github.com/mitghi/swift-algorithm-club/tree/master/AVL%20Tree]. I am completing the README file and as soon as its finished, i make a pull request.

Thankd

PR #17, opened 2016-02-03
Select minimum maximum

chris-pilcher:
I have added documentation and an implementation to select the minimum/maximum item in an array.

Let me know any comments/things I could change/improve!


hollance:
Looks great. Thanks!

Are you aware of the algorithm that lets you find both the minimum and maximum at the same time but doing only 3 comparisons for every 2 items?

That would be a cool thing to add, if you feel so inclined. :-)

chris-pilcher:
Thanks! No I haven't seen that algorithm. Do you have a link/book recommendation that explains the algorithm?

hollance:
I've seen it explained in Elements of Programming Interviews. 

This page also has it (method 3), but the code looks more complicated than it needs to be: http://www.geeksforgeeks.org/maximum-and-minimum-in-an-array/

PR #18, opened 2016-02-03
Just a small update on how to get midIndex

avoxm:


hollance:
Why this change? The article actually explains that the change you're suggestion contains a subtle bug.

avoxm:
Hmmm ... interesting that you said that it contains a bug, since there was no functional change it was only "syntactic" change... Why ? Let's see the code :)

Initial :  range.startIndex + (range.endIndex - range.startIndex) / 2
Purposed : (range.endIndex + range.startIndex) / 2

Let's do the math. I am saying that 
range.startIndex + (range.endIndex - range.startIndex) / 2 == (range.endIndex + range.startIndex) / 2 

Proof : 
let range.startIndex = 0
let range.endIndex = 16

Then : 

0 + (16 - 0) / 2 should be equal (16 + 0) / 2
Which is the case.

Please let me if there's something I don't see right here :)

hollance:
Let's say we're working with unsigned 32-bit numbers and my array contains 4.294.967.296 elements. Now at some point the startIndex will be > 0, for example 10. What happens when you add 10 to 4.294.967.296 when you're using 32-bit numbers? ;-)

avoxm:
I see what you mean, it is a valid point, still to be honest don;t think it is a valid scenario :)

hollance:
It doesn't really matter what the actual limit is. Even with an Int32 you can cross that limit.

PR #20, opened 2016-02-04
add O(n log n) sol for kth largest element

D-Speiser:


hollance:
Nice! Are you aware of the O(n) solution that combines ideas from binary search and quicksort? I had this on my to-do list already but if you want to add it that's fine too.

D-Speiser:
I've seen O(n) solutions using heaps or quick select, although I have never implemented them myself. Let me know if you'd like me to add one of those solutions, and I'll get around to it when I can. Otherwise I was thinking of working on the machine learning section, since that seems to need the most work.

hollance:
Feel free to work on whatever interests you most. :-)

I will add the O(n) solution myself then, since I already have that lying around.

PR #21, opened 2016-02-04
Implementing Binary Tree

dev-hartmann:


hollance:
Hi. Are you working on a general-purpose binary tree or a binary *search* tree? We've already got a binary search tree so you may want to look at that (and possibly augment that explanation).

dev-hartmann:
Hi, seems like I was to quick on the "let's implement that" - side ;) I'm unsure if implementing a general purpose binary tree makes a lot of sense, because I think a binary tree is an abstract data structure and you would use one of the specialized binary tree forms like a BST or a binary Heap. Maybe I could Implement Red-Black tree. What do you think ?

hollance:
I had already planned a quick article on a general purpose binary tree, so I might as well do that myself.

But if you want to implement a red-black tree (and explain how it works), that would be great! :-)

PR #22, opened 2016-02-04
Select min and max but doing only 3 comparisons for every 2 items

chris-pilcher:
Thanks for the link to http://www.geeksforgeeks.org/maximum-and-minimum-in-an-array/ :smiley:

I've added my implementation in Swift. Let me know if you see something I can improve/change.


hollance:
Sweet! :-)

hollance:
By the way, please add your name to the bottom of the README. Something like *Written by My Name*, so you get credit for your work.

PR #24, opened 2016-02-04
AVL Tree

mitghi:
As i mentioned in previous pull request, I rewrote the original Binary Tree to AVL tree. I appreciate if somebody helps with README, but i will improve it for sure.


hollance:
Nice. It will be good to have an AVL tree and yours seems very complete.

Because the code is so big, I'd like to do some code review first, to make sure it's as good as it can be. People are going to learn from this, so we should set a good example. :-)

Here are some of my remarks:

Value and key:

- I'm not sure why it's useful to have a `key` and a `value` field. It's not necessarily wrong, but I'd just use `value` and have the key be a part of that value object.
- If you insist on keeping a separate key and value, I suggest we use `Value` and `Key` as the type names instead of `A` and `B`.
- In that case I also think it's better if `value` has the type `B?` instead of `B!`. Proper optionals are a lot safer than implicitly unwrapped ones.

Naming:

- You often write things like `leftchild`. It's more Swifty to write `leftChild`.
- Likewise for `is_leftchild`, which should be `isLeftChild`.
- Use `public` and `private` to make a distinction between methods that should be hidden. Now you often use an underscore but that's not the Swift way.

Other stuff:

- Functions like `isroot()` might as well be properties: `var isRoot: Bool { ... }`.
- You write things like `if self.parent != nil`. If more Swifty to use `if let parent = self.parent { ... }`. That's what `if let` is there for. :-)
- Personally, I don't like the use of `self.` everywhere, but that's up to you. (I think it clutters up the code.)
- You have 3 init methods that are all very similar. It would be nicer to make two of them `convenience` inits that just call the main one. Like `init(key: A) { init(key: key, value: nil, leftchild: nil, ... etc) }`.
- I'm not sure what `spliceout()` is supposed to do. It's not a term I recognize.
- `_delete()` has an optional as parameter and then you write `item!` everywhere in the method. So there's no real reason to make it an optional.

A long list but I hope this isn't too discouraging. ;-) Let me know how you get along.

mitghi:
Thanks a lot for the time you've spent on reviewing and writing these remarks, I very appreciate it.
You are absolutely right, i take these remarks into consideration and will submit it via another pull request.

I have a few notes:

1. I too think value could be confusing as well, so it would be better to use `payload` instead, because each node can have a payload which differs from the key used for ordering.
2. I think using self helps by explicitly showing the reader that the variable is an instance property.
3. `spliceout()` is similar to the logic in `_delete()`. It detaches the leftmost child of right child from node where node is intended to be removed and has two children. This is done by using `find_successor()` and then a call to `spliceout()`. It updates the connections accordingly.

hollance:
What I meant with key/value is that the key is often part of the payload. You don't need to make the distinction. In Swift we often use an `isOrderedBefore` closure that would compare two payloads based on their key (whatever that may be). For example with sorting.

mitghi:
I understand the point. Please take a look at the code, i fixed some issues and improved the _delete function ( It didn't update the balance factor on removal ) . 

    var test = AVLTree<Int,String>()
    test[10] = "test_10"
    test[11] = "test_11"

    test.print_values()
    print("")

    test[12] = "test_12"

    test.print_values()
    print("")

    test.delete(11)

    test.print_values()
    print("")

    test[13] = "test_13"
    test[14] = "test_14"
    test[16] = "test_16"

    test.print_values()
    print("")
    test.delete(14)

    print("")

    test[14] = "test_14"

There are a couple of issues still left and I need your help and advice. 
First, for some reason convenience init is not working, and that was the reason why i wrote the additional inits not as a convenience . I looked around on the Internet and seems that it might be a bug from compiler. If you can please check it with convenience init . The closest solution i found is using extension or a factory function.

hollance:
I'm curious why `Payload` needs to be `Comparable`?

This one works:
```
convenience init(key: Key,value: Payload){
    self.init(key: key, value: value, leftChild: nil, rightChild: nil, parent: nil)
}
```

This also works:

```
  convenience init(key: Key){
    self.init(key: key, value: nil, leftChild: nil, rightChild: nil, parent: nil)
  }
```

but you need to change the main init to:

```
  init(key: Key, value: Payload?, leftChild: TreeNode<Key,Payload>?, rightChild: TreeNode<Key,Payload>?, parent: TreeNode<Key,Payload>?){
```

so that value is now an optional.

mitghi:
Oh I see, only Key has to be Comparable, my mistake.
Now i see the error in init, thank you. 
I committed these changes.

hollance:
Thanks. Is this done now as far as you're concerned or did you want to make more changes? Let me know if you're done, then I'll merge it and write an explanation.

mitghi:
Yes this is done.

PR #25, opened 2016-02-05
Added tests for select minimum maximum algorithms

chris-pilcher:


hollance:
Thanks! Something I like to do with these kinds of test cases is add randomized tests. So you'd create a random array of a certain size and then test your own code and make sure it gives the same result as Swift's built-in code (or by sorting the array).

In pseudocode:

```
for _ in 0...10 {
  for n in 0...100 {
    create random array of size n
    calculate result
    assert that it is corret
  }
}
```

chris-pilcher:
Thanks for the tip! I'll add that test right now. 

Do you know if XCTest contains something similar to the NUnit repeat attribute? I had a quick google but I couldn't find anything.

PR #27, opened 2016-02-05
Checking algorithm against Swift library algorithm

chris-pilcher:
Added tests that compare the algorithm against the Swift library algorithm.

Also fixed some typos in the test names.


hollance:
You're on a roll, Chris! :-)

PR #28, opened 2016-02-05
Handling empty lists when selecting minimum maximum

chris-pilcher:
I've updated the algorithms to handle empty lists. This means the algorithm will return nil when given an empty list which matches the behaviour of the Swift library.

Do you think this should be included? It means we have to add a guard to each algorithm which increases the complexity. I'm not 100% sure if this pull request should be included.


hollance:
It's good practice, so we might as well include it.

PR #30, opened 2016-02-06
Priority Queue

mitghi:
This is a Priority Queue ( min priority oriented ) based on min-heap properties.
I did not used Generics because of your note on README file, and as well droped using `self.` in instance methods.

Example:

    var test = PriorityQueue()
    test.insert(19)
    test.insert(2)
    test.insert(2)
    test.insert(4)
    test.insert(1)
    print(test.pop())
    print(test.pop())
    print(test.pop())
    print(test.pop())
    print(test.pop())
    print(test.pop())
    print(test.pop())
    print(test.pop())

Thanks.


hollance:
Thanks. I'll look at this later. :-)

mitghi:
I also added Shell sort.

hollance:
Could you add shell sort in a different pull request? (You may need to make a new branch for that.)

I need some time to review the priority queue because we already have a heap and I want to review that one in more detail too. But the shell short looks like it can be merged sooner than that, so it will be helpful to have that in its own pull request.

Note: Your shell sort does `for i = start + gap; i < list.count; i += gap {`. This type of for loop is going away in Swift 3.0. I suggest you change it to something like `for i in (start.gap).stride(to: list.count, by: gap)`.

hollance:
I've decided not to use this priority queue. It seems fine but since we already have a heap, I decided to build it on top of that one instead. Thanks anyway!

PR #38, opened 2016-02-08
Fix segment tree

goingreen:
I found critical bug in my implementation. Now it fixed and I checked that by some performance tests. As an apology I added some more illustrations in description :smile: . Also there is mistake in main readme: it's not [that](https://en.wikipedia.org/wiki/Segment_tree) segment tree, but [this](http://wcipeg.com/wiki/Segment_tree)


hollance:
Cool. I will probably merge it tomorrow or the day after (busy right now).

PR #40, opened 2016-02-09
Union-Find data structure implementation

goingreen:
I have only to write an explanation. I think it will be ready tomorrow.


hollance:
What does the DS stand for?

goingreen:
It's abbreviation for DataStructure. Do you think that it's redundant?

hollance:
Yeah, I don't think "DS" adds anything so I would leave it out.

Have you considered making it a `struct` instead of a `class`? In Swift most collection types are structs.

goingreen:
I haven't thought of it as collection type. But I've taken into consideration your сomment and now changed my mind so I suppose to make it a `struct`.

goingreen:
I want to refactor this very unsafe line of code `let indexOfElement = index[element]!`. But I don't know what would be better: calling `fatalError()` or throwing exception? @hollance What are your thoughts?

hollance:
I would solve this by returning nil if the element cannot be found:

```
public mutating func findSetOfElement(element: T) -> Int? {
  if let indexOfElement = index[element] {
    return findSetByIndexOfElement(indexOfElement)
  } else {
    return nil
  }
}
```

hollance:
By the way, can you rename the directory to `Union-Find`?

goingreen:
Also I'm thinking about helper method like
```swift
public mutating func isInTheSameSet(firstElement: T, secondElement: T) -> Bool {
  if let firstSet = findSetOfElement(firstElement), secondSet = findSetOfElement(secondElement) {
    return firstSet == secondSet
  } else {
    return false
  }
}
```

hollance:
I have some suggestions for name changes. Not that your names are bad, but they could be simplified a little.

- `isInTheSameSet()` could just be `inSameSet()`
- `findSetOfElement()` could be `setOfElement()` or even just `setOf()`
- `addSetWithElement()` could just be `add()`
- `unionSetsWithElement(andSecondElement:)` could be `unionSetsContaining(and:)`

I also notice you used `( )` with some if-statements and even some semicolons! :scream: ;-)

goingreen:
I still write some ObjC but I'm honestly trying to get rid of this habit :flushed:
I have some doubts about renaming `addSetWithElement()` to `add()` because it becomes a bit not clear what we actually add. In my opinion, the best solution is `addSetWith()`

hollance:
The idea is that each element you add is given its own set containing just that element, right? I agree that just `add()` may not make that very clear.

goingreen:
I think I've done. Do you have any remarks?

hollance:
Looking good. :-) I'll probably merge it tomorrow.

hollance:
Very nice! I tweaked the description a little for grammar etc.

goingreen:
Thank you for your work. Now it looks much better!

PR #41, opened 2016-02-10
Added fizz buzz

chris-pilcher:
Added fizz buzz puzzle to list of interview puzzles.


hollance:
I feel a bit silly about including this in the repo, but you did a great job on the write-up so I feel guilty about not including it. ;-)

Could you make a small change? You have the results listed as `True` or `False`. This might be confusing as they aren't written with a capital letter in Swift. So `true` and `false` would be better.

chris-pilcher:
I was unsure if it should be included. :smile: 

It's definitely a software developer interview question but it's not really an *algorithmic* puzzle.

hollance:
Well, I'll merge it anyway. :-) As punishment I will now give you the task of writing about a very difficult algorithm, he he. ;-)

PR #26, opened 2016-02-05
Graph Implementation

donald-pinckney:
Working on the plain graph code, not yet close to done.  Currently I am focusing on representing directed, weighted graphs using adjacency lists.  Feedback is welcome!

PR #42, opened 2016-02-11
Breadth first search

chris-pilcher:
I'd like to work on the breadth-first search algorithm :smile: I should have it done in the next couple of days.


hollance:
I assume you're going to do this on a graph? Note that someone is already working on graphs in general (see https://github.com/hollance/swift-algorithm-club/pull/26 ), which may be useful to know. ;-)

chris-pilcher:
I am planning on using a graph. Thanks for letting me know about #26.

hollance:
I don't think it's necessary to explain the Graph and Queue etc in too much detail. Just explain why you need these things and link to the existing explanations we have for them.

The focus of this article should be on BFS: how it works and what you can do with it. 

If people want to see exactly how you implemented Graph, Node, Queue etc they can look in the source code. No need to show their entire code in the README, that just distracts from the topic at hand.

chris-pilcher:
Thanks for the input! :+1:  I'll reduce the details about graphs and queues and link to existing explanations. I'll add examples of what you can do with BFS.

hollance:
Some applications of BFS are:

- Use BFS to compute the shortest path between a source vertex and each of the other vertices. (Unweighted graphs only.)
- Calculate minimum spanning tree on an unweighted graph.

hollance:
Nitpicking here, but if you write:

```
let current = q.dequeue()
for edge in current!.neighbors {
```

then I think the following is better:

```
let current = q.dequeue()!
for edge in current.neighbors {
```

otherwise you have to unwrap current everywhere you want to use it. Now the unwrap happens in just one place and it's clearer why you need to unwrap it in the first place.

hollance:
@chris-pilcher Did you want to do more work on this, or can I merge it already?

chris-pilcher:
I'd like to do more work on this. I will work on it tonight. Thanks!

hollance:
Cool, just let me know when it's ready to go.

chris-pilcher:
I've updated the readme + playground for the minimum spanning tree example.

The only thing left to do is to add a test project. I will do this tomorrow night.

hollance:
Awesome, this looks great.

hollance:
I'll already merge it.

PR #43, opened 2016-02-11
change null-coalescing to nil-coalescing

drew-y:
[From the docs](https://developer.apple.com/library/ios/documentation/Swift/Conceptual/Swift_Programming_Language/BasicOperators.html#//apple_ref/doc/uid/TP40014097-CH6-ID72)


hollance:
Good catch!

PR #45, opened 2016-02-18
Treap using Indirect Enums

rothomp3:
So, this is not pretty and does not (yet) contain much in the way of teachable moments or explanations, but here it is anyway. Bonus includes conformance to `CollectionType`!

### DIFF
--- a/Treap/Treap.swift
+++ b/Treap/Treap.swift
@@ -1,0 +1,160 @@
+//
+//  Treap.swift
+//  TreapExample
+//
+//  Created by Robert Thompson on 7/27/15.
+//  Copyright © 2015 Robert Thompson. All rights reserved.
+//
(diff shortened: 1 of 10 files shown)


hollance:
Hey, this looks pretty good. You mentioned you won't have time to write an explanation, but I think we can use this code in one form or another. :-) Thanks!

rothomp3:
I'm on my iPad right at the moment, but as soon as I get back to my desk I'll fix the license and update the pull request. Total brainfart on my part leaving that "All rights reserved" line in there haha.

chris-pilcher:
I'll merge this now and add it to our new [Under Construction](../blob/master/Under Construction.markdown) area. Thanks for the contribution @rothomp3.

PR #46, opened 2016-02-19
Unit tests for Array2D and AVLTree data structures

barbaramartina:
Hi @hollance . I'm Barbara and I'd like to do a contribution to the project.
I've added 2 XCode project with its respective tests targets for _Array2D_ and _AVLTree_.
I've include some functional and performance tests.
Not all the possible tests are covered, but I enabled code coverage data so more tests could be added in the future.
There is also a small bug fix after some of the tests failed, for the class _AVLTree_: the parent property was not set on children when creating a parent node.

I hope these changes help and I will continue adding more tests, in upcoming contributions, for these structures and the rest of the project. I really like this initiative, I find it useful in many sense, to learn, to practice, to improve and generate re-usable algorithms and data structures. Thanks for creating the project.

While working on that, I was told about a website called, [**Rosetta**](http://rosettacode.org). Have you heard before about this?. I think it might be interesting to take some algorithms from there, solve them in this repo and upload the solutions to Rosetta, they have around 781 problems solved and support 605 programming languages, including **Swift**.


hollance:
Thanks, Barbara, this is great! Unit tests are always welcome. :-)

I have heard of http://rosettacode.org indeed. I shall add a link to the main page.

hollance:
I am not sure how useful the Array2D test `testPerformanceOnLargeArray()` is. It takes forever to run but all it tests is how fast `print()` works. ;-)

Maybe there is a better way to test this...

barbaramartina:
@hollance yes you're right, and it takes for ever... I really don't know the performance of print()... I'll think about a better way to test it removing the calls to print, maybe having only the creation with several items... to test only the initialization time.

hollance:
Here's something to think about: Inside the loop, do `array[c,r] = c*r` to give it a value. Then repeat the loops, but do an assert to make sure `array[c,r]` really is `c*r`.

Then repeat this but do `array[r,c]` instead of `[c,r]`. If the array is big enough, this should be slower because of the array not completely fitting into the CPU cache.

That sounds like a useful test. :-)

hollance:
NOTE: I changed the structure of the unit tests. They are now in a "Tests" folder and consist of just the test bundle, so there is no longer an app you can run.

barbaramartina:
And how to do you know how big is the cpu cache really? To make this test more meaningful.

hollance:
That depends on the CPU. I was just curious if there was a detectable difference in looping by rows or looping by columns (there should be).

barbaramartina:
@hollance  about:

> 
NOTE: I changed the structure of the unit tests. They are now in a "Tests" folder and consist of just the test bundle, so there is no longer an app you can run.

I think it would be more practical in order to run the tests quickly to keep an app that we can open and run directly.

hollance:
You can open the Xcode project and press Cmd+U to run the tests.

I don't want to put all the tests into a single project because we will have naming conflicts between different algorithms and data structures. So it's better to keep the tests separate.

barbaramartina:
So Great! I have not noted the project can be run directly. I just wanted a way to run quickly the tests

PR #48, opened 2016-02-24
Implement K Nearest Neighbors

johnsgill3:
Todos:
- [ ] Write algorithm
- [ ] Write README


hollance:
Nice! 

I took a quick look through the code and my eye caught this:

```
anotherIndex != index ? swap(&elements[index], &elements[anotherIndex]) : ()
```

That seems like an abuse of the `?:` operator. ;-) I'd prefer it if you wrote that using an if-statement, or alternatively use this swap function that I've used elsewhere too:

```
public func swap<T>(inout a: [T], _ i: Int, _ j: Int) {
  if i != j {
    swap(&a[i], &a[j])
  }
}
```

Likewise, you may want to create a helper function for returning random numbers instead of calling arc4random directly:

```
public func random(min min: Int, max: Int) -> Int {
  assert(min < max)
  return min + Int(arc4random_uniform(UInt32(max - min + 1)))
}
```

It makes things a little easier to read.

Looking forward to the README! :-)

hollance:
Oh, am I correct in understanding that you use `Vector2` for the data points? If yes, how will you handle data that has more features? It might be useful to make the vector class more general.

johnsgill3:
Plan is to implement a much cleaner random index algorithm. Wasn't 100% happy with this one from a performance standpoint and it's a little complex. But for the naive simple first pass I was happy with it.

As for the `Vector2D` I will see what I can do to make it work for N Dimensional objects.

Hope to get the REDME completed by weekend.

johnsgill3:
Need to re-do this code as I totally had mind blank and implemented k-means instead of kNN. Moved the code into new branch and opened new PR. Will get this one finished after I write README for K-Means.

chris-pilcher:
Hi @johnsgill3. Just checking in. Are you still interested in implementing K Nearest Neighbors?

kelvinlauKL:
Doing a cleanup of inactive PRs. If you're still interested in writing this, feel free to remake the PR!

PR #50, opened 2016-02-25
Use the ternary operator for code clarity and one line consistency

rajabishek:
Uses the ternary operator to check the length of the array and return a nil or the actual removed value from the method. Changes the code to add better clarity and maintain the single line code consistency in the stack structure

### DIFF
--- a/Stack/README.markdown
+++ b/Stack/README.markdown
@@ -57,11 +57,7 @@ public struct Stack<T> {
   }
 
   public mutating func pop() -> T? {
-    if isEmpty {
-      return nil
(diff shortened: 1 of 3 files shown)


hollance:
I understand why you're making these changes but this is not a pull request I'm willing to accept. :-(

In my own personal code I'd probably use the `?:` too in such situations. However, the main purpose of this repo is to explain things to people, and I feel that using shorthand notations such as `?:` make things less obvious, not more.

rajabishek:
@hollance I understand.

PR #51, opened 2016-02-26
Implement K-Means algorithm

johnsgill3:
An Unsupervised classification machine learning algorithm.

Totally blanked on kNN vs kMeans and implemented the wrong one. So rather than waste code, moved it over to a new branch and made a few of the comments from PR #48.

Opening pull request to get feedback on code.

Todos:
- [x] Write README.md


hollance:
Hey, more algorithms is better! :-D

johnsgill3:
@hollance Can I get a review of code and README for potential merge?

hollance:
Some minor points: 

I would call `VectorND` just `Vector`. I also wouldn't call the methods `getData()` and `getLength()`. That's what Java people do. ;-) Instead, you can declare the `data` and `length` properties as `private(set) public`. That way you don't even need the methods.

I prefer to use a separate `random()` function instead of using `arc4random_uniform()` directly, since it can be hard to figure out exactly what range `arc4random_uniform()` works on.

It's best not to use `++` since this is going away in Swift 3.0. `+= 1` makes the code future-compatible.

In the playground, when the code becomes large I usually add it as a separate source file. That keeps the actual playground cleaner.

johnsgill3:
I changed Vector into it's own source file. It could be expanded upon and maybe even moved to it's own directory as a Data Structure.

I'm still learning swift which is why I found this repo to be enjoyable. Refresh myself on algorithms and implement in new language. Still learning some of the nuances of swift over other OO languages.

Made change to use random() and modulus, as well as removed all '++'.

Changed code from a playground to a project with a few tests.

hollance:
Great, I'll look into this soon. I agree that Vector deserves a topic of its own. I've been writing my own Matrix class for doing things like logistic regression. It uses the Accelerate framework to make things really fast. Maybe we could do something similar for Vector.

hollance:
OK, I finally had a look at it (and edited the text and code a little). It would be great if there was a bit more explanation of how the code actually works. The illustrations are very useful already, but a step-by-step explanation of the logic would also be good to have. :-)

johnsgill3:
@hollance wrote Issue to track it.

PR #58, opened 2016-02-29
Breadth first search tests

chris-pilcher:
I have added tests for BFS example, BFS shortest path and BFS minimum spanning tree.

In order to add these tests I had to change the example functions. Here are a summary of the changes:
- Made simple example return a list of explored nodes
  - Previously it was printing explored nodes inside of function
- Made shortest path return a shortest path graph
  - Previously it was printing shortest paths inside of function
- Implemented equality on edges, graphs and nodes to allow for tidy asserts
- Changed simple example to used node.visited flag instead of storing separate "seenNodes" list


hollance:
I made a few small tweaks to the text, but other than that I'm very happy with it. Thanks again!

chris-pilcher:
You're welcome! Thank you for all of the tips and advice! :smile:

PR #60, opened 2016-02-29
Update README.markdown

jherna37:
My native programming language is Java and I am new to swift.  To the best of my knowledge I have translated from Java to Swift. Thanks for the opportunity and don't be afraid to hurt my feelings. I'm open to any feedback. This was merely the first step of many.

### DIFF
--- a/Bubble Sort/README.markdown
+++ b/Bubble Sort/README.markdown
@@ -5,27 +5,22 @@ Goal: To sort an array from the lowest to highest value (or highest to lowest).
 You are given an array of unsorted values. The BubbleSort algorithm works as follows:
 
 ```swift
-   for var j = 0; j < array.count; j++ 
+   for in j = 0; j < array.count; j++ 
       {
-         for var i = 0; i < array.count; i++ 
+         for in i = 0; 0 < array.count - 1; i++ 
          {
-            var current = array[i]
             
-            if i + 1 < array.count
-            {
                if current > array[i + 1] 
                {
-                  array[i] = array[i + 1]
-                  
-                  array[i + 1] = current
+                  swap(&array[i], &array[j])
                }
-            }
+            
          }
       }
 ```
 This code can be tested with the following:
 ```swift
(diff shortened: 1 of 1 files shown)


hollance:
Swift 3.0 is going to come out later this year and it removes this kind of for loop:

`for var j = 0; j < array.count; j++`

I also think you can get rid of the `if i + 1 < count` line if you make the loop for `i` go from `0` to `count-1`.

You can swap two array elements in Swift with: `swap(&a[i], &a[j])`, so no need to use a temporary variable.

It is possible to make the bubble sort a little bit more efficient: if no swap was necessary, then the sort is done and you can exit the loop.

That should give you something to do. ;-)

hollance:
@jherna37: Is this something you're willing to work on some more?

jherna37:
Hey sorry for the delay. I just finished my mid term exams and haven't had
much time since my first submission to work on it.

hollance:
No worries. I'll just leave this open until you get a chance to work on it again.

jherna37:
Just for clarification, would you like me to implement this algorithm without the use of for loops?

hollance:
You can still use for loops, but it must be `for in`, not `for var i = ...; ... ; ...` since the latter is now deprecated in Swift.

jherna37:
I've made the changes to the code. I was wondering if instead of radix sort you'd be open to more algorithm interview questions. I have a few examples from an interview I just came back from and think they'll make a nice addition.

hollance:
Have you actually tried out those changes? I'm pretty sure it doesn't work like that...

Interview questions are very welcome! We already have a few simple ones, but more is better. :-)

chris-pilcher:
Hi @jherna37. We're in the process of cleaning up our pull requests.

Would you still like to work on the bubble sort algorithm?

kelvinlauKL:
Closing this due to inactivity. @jherna37 feel free to open another PR if you're up for another shot!

PR #63, opened 2016-03-01
Adding simple example for Depth-First Search

paulot:
This adds a simple example with an explanation of a DFS. I did not create animations yet for this example, as I'm not very good with animations or image handling. The example used the same format as used for the BFS example, please let me know if there are any improvements you want to see.


hollance:
Thanks! I'll merge this and then edit it a little.

PR #64, opened 2016-03-02
Bounded priority queue

johnsgill3:
In working on kNN implementation need to implement KDTrees and needed to implement a special type of Priority Queue that is bounded. Moved Bounded Priority Queue to it's own branch as it's an interesting container.

I changed the implementation from a heap (normal Priority Queue) to use a LinkedList.


johnsgill3:
@hollance Any comments on code or README?

hollance:
Thanks! I'll merge this and make some (small) edits.

hollance:
Actually, I made quite a few changes to the code. The structure is still largely the same but I renamed some variables and simplified some of the if statements.

I'm not 100% happy with the README yet, but it's a start. :-)

PR #66, opened 2016-03-02
More test cases for AVLTree

barbaramartina:
Tests to cover:
- minimum node
- successor of a node
- delete a node given a key


hollance:
Thanks!

If you have some time, would you mind making a unit test for this bug that has been reported? https://github.com/hollance/swift-algorithm-club/issues/61

barbaramartina:
Hi! Sure. Might next week I'll have a couple of hours and will check that bug.

PR #68, opened 2016-03-03
Topological Sort

aliHafizji:
Just wanted to book this algorithm so creating a PR. 

I'll have the code up in a few days.


hollance:
Nice. Are you going to use a depth-first search for this?

aliHafizji:
@hollance Yeah I was planning on using DFS. Since, it's already part of this project and explained I was just planning on reusing that and explaining the idea behind topological sort.

hollance:
Sure. But it's OK if you want to use your own implementation. There are many different ways to implement graphs and DFS and BFS so it's fine if not everything uses the exact same classes etc.

aliHafizji:
@hollance I think this is ready for review. Let me know what you think.

hollance:
Cool. I'll look it over later today. There's an alternative algorithm for topological sort that I want to add that does not use a DFS. ;-)

hollance:
Hey Ali, I think your topological sort code is incorrect. If you look at the image I added (at the bottom of the README), you'll see that some of the arrows go the wrong way.

I also added the alternative algorithm (in TopologicalSort2.swift) and as you can see, it does give a different sort order. It's possible to have many different topological sorts for a single graph, but I'm afraid yours is not correct. For example, 11 can never happen before 5 since it depends on 5.

hollance:
@aliHafizji FYI: I've worked on this some more. The bug was that the results were added into the sorted array in the wrong order. Also during the DFS. But it should be fixed now. :-)

PR #69, opened 2016-03-04
Added a Array2D.playground  

Thecmar7:
Added a playground for the Array2D


hollance:
Thanks! I renamed the variable to `matrix`.

PR #70, opened 2016-03-09
no need to set neighbor as it's already set in constructor

infynyxx:
Minor change


hollance:
Thanks! I was just making some changes to that stuff, so I've also incorporated this improvement. :-)

PR #76, opened 2016-03-13
Introduce Associated Labels to centroids.

perlfly:
This could help showing how to use KMeans as a classifier.

Some changes to KMeans properties and function.
Updated Tests accordingly.
Demo the usage of labels in one small test.

### DIFF
--- a/K-Means/KMeans.swift
+++ b/K-Means/KMeans.swift
@@ -5,21 +5,24 @@
 
 import Foundation
 
-class KMeans {
+class KMeans<Label: Hashable> {
   let numCenters: Int
-  let convergeDist: Double
+  let labels: Array<Label>
+  private(set) var centroids: Array<Vector>
 
-  init(numCenters: Int, convergeDist: Double) {
-    self.numCenters = numCenters
-    self.convergeDist = convergeDist
+  init(labels: Array<Label>) {
+    assert(labels.count > 1, "Exception: KMeans with less than 2 centers.")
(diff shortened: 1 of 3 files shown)


hollance:
Thanks! I'm going to leave this here for the minute. I still want to look over the k-means explanation and code. Yours looks like a good improvement :-) but it might be a few days before I get around to it.

perlfly:
Thank you @hollance . Consider it as a humble proposal. Here I add `Label` as a convenience way to show the results. K-Means is often used in machine learning as an unsupervised learning algorithm. You _train_ the centroids using some training points. Then you use the trained classifier in order to classify previously _unseen_ points. This is why I use `train` and `fit` in function names.

I see the code has almost no comment. Thus I did the same, since I don't know if this is functional to the project intentions. Maybe explanations are meant to go in the readme.

hollance:
This repo is all about explaining stuff. Some source code in here doesn't use a lot of comments because it's not that hard to follow what's going on. But in general I like to err on the side of too many comments rather than too few.

I have a note on my to-do list that the README for k-means could do with more depth and an explanation of how the code actually works. If you're up for that, it would be a very welcome contribution. :-)

hollance:
@perlfly: All right, I finally had some time to look at this. Merged and looked at. :-) Thanks again!

perlfly:
@hollance  My pleasure! I hope I'll be able to do some more.

PR #77, opened 2016-03-13
Refactor linked list

andela-aonawale:
Refactor reverse method of LinedList class to make the code more Swift-like, without the forced unwrapping.


hollance:
Hey, that's great! Much more Swift-like indeed. :-) Thanks!

PR #86, opened 2016-03-20
Fixed the count condition 

hamdullahshah:
Make it same as explanation, i.e at least 50 elements and 0.25 percent.


hollance:
Thanks!

PR #87, opened 2016-03-22
Implementing Ordered Set

zainh96:
Hey, I've decided to write the Ordered Set class. I have a decent amount of code for it already, is there anything in particular (function, feature, topic etc) you would like specifically for this data type?

It shouldn't take me too long to complete this, maybe a day or two. 

Thanks! :)


zainh96:
Forgot to add something. Let me make a different pull request

PR #88, opened 2016-03-22
Implementing Ordered Set

zainh96:


zainh96:
Hey, I've decided to write the Ordered Set class. I have a decent amount of code for it already, is there anything in particular (function, feature, topic etc) you would like specifically for this data type?

It shouldn't take me too long to complete this, maybe a day or two.

Thanks! :)

hollance:
Cool. I think it's more important to have a clear description than a ton of code, so as long as the code does the basics of what an ordered set needs to do, I'm happy. :-)

zainh96:
I've finished the section on Ordered Sets. Let me know if there's anything you'd like me to change :).

hollance:
Cool! I've edited the text and code a little bit for clarity. Thanks for your contribution! :-)

Some points:

- I'm not sure I see that the total performance of insert() is O(n log n). Inserting in an array is O(n) and exists() is O(log n + k), so if I'm doing my math correctly the total time would be O(n + log n + k) = O(n).

- Likewise for remove().

PR #90, opened 2016-03-25
Add a faster approach to calculate C(n, k)

nuts23:
Add a faster approach to calculate `C(n, k)` in `O(k) time and `O(1)`
extra space.


hollance:
Thanks! That's a useful improvement. :-)

PR #91, opened 2016-03-27
Fixed Complexity for insert and remove functions

zainh96:
It was pointed out to me that the insert and remove functions on arrays have complexity of O(n). I've changed the readme and code to reflect that.


hollance:
Ah yes, I noticed that too. (See my comment on your pull request.) Since I've already made a bunch of changes, I'll just fix that myself right now.

zainh96:
Perfect, thanks!

PR #92, opened 2016-03-27
Fixing deallocation of LinkedListNode

rinat-gabdullin:
Using of strong references for both `next` and `previous` properties leads to retain cycle. So `LinkedList.removeAll()` doesn't lead to deallocation of `LinkedListNode` objects.

### DIFF
--- a/Linked List/LinkedList.swift
+++ b/Linked List/LinkedList.swift
@@ -6,7 +6,7 @@
 public class LinkedListNode<T> {
   var value: T
   var next: LinkedListNode?
-  var previous: LinkedListNode?
+  weak var previous: LinkedListNode?
   
   public init(value: T) {
     self.value = value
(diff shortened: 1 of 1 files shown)


hollance:
Hey, that makes a lot of sense. I hadn't considered ownership cycles.

But I have a question: why the change in reverse()?

rinat-gabdullin:
Hi Matthijs.

Because after `swap` node in `previous` becomes unretained. Node is being freed and test fails. So we need to retain it in `node` local variable before swapping.

hollance:
Ah yes, that makes sense. :-) I'll merge this soon.

rinat-gabdullin:
Thanks. It's my first public PR ever :-)

hollance:
OK, it's in there. Thanks very much!

PR #93, opened 2016-03-29
Improve Boyer-Moore by always using skipTable to advance cursor

inamiy:
```swift
// If no match, we can only safely skip one character ahead.		
// The characters are not equal, so skip ahead. The amount to skip is
i = i.successor()
```

I think this is inefficient and you can always use `skipTable` to skip many characters.

### DIFF
--- a/Boyer-Moore/BoyerMoore.playground/Contents.swift
+++ b/Boyer-Moore/BoyerMoore.playground/Contents.swift
@@ -5,41 +5,38 @@ extension String {
     let patternLength = pattern.characters.count
     assert(patternLength > 0)
     assert(patternLength <= self.characters.count)
-    
+
     var skipTable = [Character: Int]()
     for (i, c) in pattern.characters.enumerate() {
       skipTable[c] = patternLength - i - 1
     }
-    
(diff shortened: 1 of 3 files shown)


hollance:
Can you explain this change a bit more?

inamiy:
Sorry, I discovered that my implementation had potential infinite-loop bug due to backward sliding by bad-character-heuristic (e.g. `"abcacabcab".indexOf("abcab")` ).

What I was familiar with this algorithm was rather a revised one called [Boyer-Moore-Horspool algorithm](https://en.wikipedia.org/wiki/Boyer%E2%80%93Moore%E2%80%93Horspool_algorithm).
This seems to be a better example to demonstrate, but I'm not sure if it fits to this project.

I wrote a gist here https://gist.github.com/inamiy/f4e709d2838a3c230adf0cadf0daacb8 .
If you are interested, I will send another pull request :wink:

hollance:
Hmm, that looks interesting. :-)

It would be cool if you could add this code in a new section at the end of the README, and explain a bit what the differences are. Thanks!

inamiy:
Please check #94 as a new PR. I will close this one.

PR #98, opened 2016-04-05
Fix typo

keitaito:
Hi swift-algorithm-club team,

I found and fixed a couple of typos in Binary Search Tree chapter.

In README.markdown, at line 431:
Before: "Finally, we can write the code the remove a node from the tree:"
After: "Finally, we can write the code that removes a node from the tree:"

At line 673:
Before: "A usual, ..."
After: "As usual, ..."

Please check it out when you have time. It would be great to merge this pull request if it is good. Sorry if I am wrong about these changes.

Thank you.

### DIFF
--- a/Binary Search Tree/README.markdown
+++ b/Binary Search Tree/README.markdown
@@ -428,7 +428,7 @@ We won't need it for deleting, but for completeness' sake, here is the opposite 
 
 It returns the rightmost descendent of the node. We find it by following `right` pointers until we get to the end. In the above example, the rightmost descendent of node `2` is `5`. The maximum value in the entire tree is `11`, because that is the rightmost descendent of the root node `7`.
 
-Finally, we can write the code the remove a node from the tree:
+Finally, we can write the code that removes a node from the tree:
 
 ```swift
   public func remove() -> BinarySearchTree? {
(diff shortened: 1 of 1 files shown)


keitaito:
Besides this, I have a couple of questions.

- The links to threaded binary tree and red-black tree are not available, because these chapters are not added yet. Is it okay to just leave these links?

- About enum case let keywords, I just learned from this page that I can write enum's switch case with two ways.

Add let keywords to all associated values:

    case .Node(let left, let value, let right):

Add let keyword before a specific case name:

    case let .Node(left, _, right):

Both of these are used with the enum version of Binary Search Tree in this chapter. I was wondering if it might be better to use only one syntax for consistency. I would like to hear your opinion if you don't mind. 

Thank you.

hollance:
Thanks for the PR!

The links to the chapters that are not available yet should stay. Otherwise I might forget to add them when we do add the chapters. ;-)

I'm not so worried about being 100% consistent about everything. Feel free to submit a pull request if you think that a different syntax makes things clearer. But remember the goal of this project is to explain things. I'm not interested in making the code super streamlined, unless it helps with the explanation.

PR #106, opened 2016-04-15
Minimum edit distance for strings

herrlui:
Hi! 
I implemented the usual minimum edit distance algorithm for strings using the Levenshtein distance. I also plan to extend this implementation such that the algorithm can use arbitrary distance measures (as for example keyboard lookup tables).

### DIFF
--- a/Minimum Edit Distance/MinimumEditDistance.swift
+++ b/Minimum Edit Distance/MinimumEditDistance.swift
@@ -5,18 +5,25 @@ extension String {
         let n = other.characters.count
         var matrix = [[Int]](count: m+1, repeatedValue: [Int](count: n+1, repeatedValue: 0))
         
+        
+        // initialize matrix
         for index in 1...m {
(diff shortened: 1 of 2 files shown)


hollance:
Cool. :-) Could you also add an explanation of how this works?

herrlui:
I have added a README. However, this is tentative since in a next step I will generalize the algorithm for more operations and costs.

hollance:
Thanks. I'll wait with merging until you're done. :-)

hollance:
Quick announcement: this repo is moving to https://github.com/raywenderlich/swift-algorithm-club -- everything should still work as before but some old links may not work anymore.

chris-pilcher:
Hi @herrlui. Would you still like to make more changes before this is merged? 😃

herrlui:
Hi! Sorry for the delay! I still plan to extend my code as soon as I find some time! :)

PR #138, opened 2016-06-09
Added swift-algorithm-cllub.workspace so all projects can be viewed together

kevinrandrup:
Using a workspace brings all of the projects together in a single window with all of the code highlighting, search and documentation features that Xcode brings. I found it extremely useful when viewing/comparing multiple projects.


chris-pilcher:
Thanks @kevinrandrup! 👍 This is also really helpful for finding existing issues across all projects:

image

I'll add new algorithms to this workspace when they are submitted to the repository.

PR #165, opened 2016-07-16
Added SlowSort

Dabendorf:
I have added another ineffective algorithm as sorting algorithm example.


kelvinlauKL:
Ok, merging this in :]

PR #166, opened 2016-07-17
Added Comb Sort

srutstein21:
Added the files for Comb Sort, a more time efficient version of Bubble Sort. Let me know how it looks, and if there is anything to update/change (and why preferably 👍).


srutstein21:
@kelvinlauKL Can you check this over? Let me know if there are any issues.

kelvinlauKL:
@srutstein21 Sorry for the wait, I've went over the PR!

Did you know Swift already has a `swap` method introduced at the global scope? I've removed yours since they're essentially the same. Let me know if I should revert it if you had another reason for it. 

Thanks for the contribution! I've merged this.

PR #176, opened 2016-08-09
Refactor insertion sort

velyan:
This is just a minor optimization to the code that takes advantage of the inout keyword.


velyan:
Closing as tests need to be reworked as well.

PR #186, opened 2016-08-21
Updated Bubble Sort description

lostatseajoshua:
## PR Changes
Added the runtime, memory and a better description of bubble sort. The implementation was not added but it is good to know the concept.

###### Discussion
Even if the algorithm is "horrible" a description of it should be added for developers to grasp the concept of bubble sort. Bubble sort is a very common sorting technique taught when beginning to learn sorting algorithms. I do believe the algorithm should be implemented and shown but that is a change for another day.

### DIFF
--- a/Bubble Sort/README.markdown
+++ b/Bubble Sort/README.markdown
@@ -1,4 +1,13 @@
 # Bubble Sort
 
-This is a horrible algorithm. There is no reason why you should have to know it.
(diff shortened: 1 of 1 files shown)


hollance:
> Bubble sort is a very common sorting technique taught when beginning to learn sorting algorithms.

Yes, and in my opinion it shouldn't be. ;-)

lostatseajoshua:
Haha yeah true, it teaches a brute force approach. That is why I just added a description of the algorithm and no implementation keeping with what was implemented before. @hollance

PR #212, opened 2016-09-13
Update Linked List to Swift 3

tuliot:
## What is this?
Swift 3 is out, and Linked List needed to be updated. 

## What did I do?
I updated the `LinkedList.swift` and `LinkedList.playground` files to conform to Swift 3.
After fixing the breaking changes, I updated the names of the functions to match the [new style](https://swift.org/documentation/api-design-guidelines/#naming). For example, 

```swift
public func nodeAtIndex(index: Int) -> Node?
```
is now
```swift
public func nodeAt(_ index: Int) -> Node?
```

>Note: I chose to change 
`removeNode(node: Node)` 
to `remove(node: Node)` and not `remove(_ node: Node)`. 
I chose this because I feel like the `node` word adds valuable context to what you're doing

Please insert allthecomments so we can figure this swift thing out together 😄 🎊 🎉


tuliot:
Need to update readme

kelvinlauKL:
Thanks @tuliot

PR #216, opened 2016-09-17
Update hash table for Swift 3.0

glennrfisher:
This pull request updates the hash table implementation to Swift 3.0.

To avoid duplicated code that must be maintained in two separate files, I deleted the `HashTable.swift` file at the folder level, along with the copy/pasted code in the Playground, in favor of a `HashTable.swift` file that is included in the Playground. This seems consistent with other algorithms, but let me know if you'd prefer something different!

I also updated the function names to be consistent with the Swift 3.0 API guidelines--most of them are copied directly from `Dictionary` for consistency. I updated the readme to reflect the change in names.


kelvinlauKL:
The naming looks good. Good call on moving the code to a separate file.

Thanks @glennrfisher!

PR #221, opened 2016-09-19
Palindromes now work with strings with even length

aliHafizji:
The code did not work for strings with even length. For e.g. "moom" even though it's a palindrome was not recognized by the function.


kelvinlauKL:
Thanks for the fix @aliHafizji!

Issue #181, opened 2016-08-18
Swift 3 Migration


chris-pilcher:
# And... we're done!

Thanks everyone for pitching in!
____

Xcode 8 and Swift 3 will be formally released soon and we need to migrate everything to [Swift 3](https://www.raywenderlich.com/135655/whats-new-swift-3). 

Want to help out with the migration? Awesome! :] 
## Please follow this process for migrating to Swift 3
- Download the Xcode 8.X
- Create a pull request to "claim" an algorithm or data structure you'd like to migrate. Just so multiple people don't work on the same thing.
- Migrate the algorithm/data structure project, playground, tests and readme to Swift 3
  - Uncomment test project in [.travis.yml](https://github.com/raywenderlich/swift-algorithm-club/blob/master/.travis.yml) so it will run on our Xcode 8 Travis CI build server
- Comment in the pull request when the changes are ready to be merged

Suggestions and feedback is welcome!
## List of algorithms to be migrated to Swift 3

`nil`

## Completed Migrations
- [x] AVL Tree #183 
- [x] All-Pairs Shortest Paths #347
- [x] Array2D #189
- [x] B-Tree #203
- [x] Binary Search Tree 
- [x] Binary Search #184
- [x] Binary Tree #213
- [x] Bit Set #194 
- [x] Bloom Filter #195
- [x] Bounded Priority Queue #218 
- [x] Boyer-Moore #248 
- [x] Breadth-First #237
- [x] Brute-Force String Search #239
- [x] Bubble Sort (no code)
- [x] Bucket Sort #281 
- [x] Comb Sort #350 
- [x] Combinatorics #270 
- [x] Count Occurrences #201
- [x] Counting Sort #235
- [x] Depth-First Search #238
- [x] Deque #303
- [x] Fixed Size Array #271 
- [x] Fizz Buzz #229
- [x] GCD #219 
- [x] Graph #267
- [x] Hash Set #276
- [x] Hash Table #216
- [x] Heap Sort #334
- [x] Heap 3945383d306ac027f779461789d6b0deddca4348
- [x] Huffman Coding #344
- [x] Insertion Sort #198 
- [x] K-Means #349
- [x] Kth Largest Element #228
- [x] Linear Regression #283 
- [x] Linear Search #200
- [x] Linked List #250 
- [x] Longest Common Subsequence #243
- [x] Merge Sort #199
- [x] Minimum Edit Distance #302 
[...]

PR #229, opened 2016-09-20
updated fizzBuzz to swift3

sendilkumarn:
This is my very first try with swift

part of #181


kelvinlauKL:
Thanks @sendilkumarn!

PR #231, opened 2016-09-21
FenwickTree

farzadshbfn:
This PL will add Fenwick Tree in swift3.0 syntax. This PL is not complete yet.

TODO: ReadMe file

### DIFF
--- a/FenwickTree/Fenwick.swift
+++ b/FenwickTree/Fenwick.swift
@@ -22,7 +22,7 @@ public class FenwickTree<T> {
 	
 	
 	/**
-	- Order: O(n)
+		Order: O(n)
 	
 	- Parameters:
 	- size: size of array (not changable)
@@ -41,7 +41,7 @@ public class FenwickTree<T> {
 	}
 	
 	/**
-	- Order: O(n * log n)
+		Order: O(n * log n)
 	
 	- Parameters:
 	- array: initial array to build fenwick tree based on them
@@ -62,7 +62,7 @@ public class FenwickTree<T> {
 	/**
 	If `to` is out of bound, method return `.zero`. results are added together using `addFunction`
 	
-	Order: O(log n)
+		Order: O(log n)
 	
 	- Parameter to: zero based index on array
 	
@@ -87,7 +87,7 @@ public class FenwickTree<T> {
 	If `from >= to` method will return `.zero`. method finds two subranges `0...from-1` and `0...to` using
 	`addFunction` and will subtract them using `subFunction`
 	
-	Order: O(log n)
+		Order: O(log n)
 	
 	- Parameters:
 	- from: zero based index on array
@@ -105,7 +105,7 @@ public class FenwickTree<T> {
 	/**
 	Adds element at given index with value using `addFunction`
 	
-	Order: O(log n)
+		Order: O(log n)
 	
 	- Parameters:
 	- itemAt: zero based index on array
@@ -126,7 +126,7 @@ extension FenwickTree where T: Comparable {
 	/**
 	Searchs for first index in fenwick tree which sum of all elements in `0...index` is less or equal to `minimumSum`
 	
-	Order: O(log n)
+		Order: O(log n)
 	
 	- Parameters:
 	- minimumSum: minimum sum to serach for
--- a/FenwickTree/Fenwick.swift
+++ b/FenwickTree/Fenwick.swift
@@ -25,11 +25,11 @@ public class FenwickTree<T> {
 		Order: O(n)
 	
 	- Parameters:
-	- size: size of array (not changable)
-	- zero: array is filled with `.zero` values at the beginning. also this value will be returned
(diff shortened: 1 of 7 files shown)


kelvinlauKL:
Looking for the readme!

farzadshbfn:
Oops, I Totally forgot, I'll add in a few days.

farzadshbfn:
Well... It's added after a decade 🚶

kelvinlauKL:
Thanks :) I'll get it reviewed.

kelvinlauKL:
@narrativium You're up for this one 👍

FenwickTree/README.markdown
```diff
@@ -0,0 +1,145 @@
...
+		var index = to + 1
+		while index > 0 {
+			sum = addFunction(sum, fenwickTree[index - 1])
+			index -= leastSignificantBit(index)
+		}
+		return sum
+	}
+```
+
+We keep summing a node's value with all of it's left-siblings, this value will be equal to cumulative sum from start of array.
```
narrativium: 'all of its left-siblings'

FenwickTree/README.markdown
```diff
@@ -0,0 +1,145 @@
...
+}
+```
+
+Each node has at max **log(n)** children (Exactly **log(LSOne(n))** children), and one parent. If we drop an index'es LSOne(Least Significant One) we will reach a node's left-sibling, and if we add an index with it's LSOne, we will reach a node's parent index.
+
+![structure](http://community.topcoder.com/i/education/binaryIndexedTrees/BITimg.gif)
+
+## Updates
+
+If an item is updated, it only affects it's parent. (and of-course this relationship is recursive i.e it's parent will affect it's grand-parent and so on...) so for updating an item, we just need to keep updating values and changing index to our parent till we reach end of an array (reaching bigger index than size of the array)
```
narrativium: 'affects _its_ parent', '_its_ parent will affect _its_ grand-parent'

FenwickTree/README.markdown
```diff
@@ -0,0 +1,145 @@
+# Fenwick Tree
+
+Fenwick Tree or Binary Indexed Tree is a data structure that can efficiently update elements and calculate prefix sums in an array of numbers.
+
+Fenwick Tree is really usefull when the array is dynamically changed. Because if there aren't many changes happening to the array, cumulative sum will be probably the best way to go.
```
narrativium: 'is really _useful_'

FenwickTree/Fenwick.swift
```diff
@@ -0,0 +1,151 @@
...
+}
+
+extension FenwickTree where T: Comparable {
+	/**
+	Searchs for first index in fenwick tree which sum of all elements in `0...index` is less or equal to `minimumSum`
+	
+		Order: O(log n)
+	
+	- Parameters:
+		- minimumSum: minimum sum to serach for
```
narrativium: 'minimum sum to _search_ for'

narrativium:
One more comment (sorry...)
`index(forMinimumSum:)` is intended to return an inclusive upper bound for the range of values which add up to at least the minimum sum. However, if the input minimum sum is very high, it might be greater than the sum of all the values in the Fenwick tree.
At the moment, the function still returns the highest index in the Fenwick tree, implying the sum of the entire tree exceeds any minimum sum.
(For example: in the playground's initialisation example with an array of size 9, sum 9, asking for `index(forMinimumSum: 10)` returns 8, implying that the sum of the range 0...8 > 10, but 9 < 10.)
Should this function be refactored to return an `Int?`, returning `nil` if the minimum sum is too large? (In the same way the `Array` function `first` returns an optional because the array might be empty.)

farzadshbfn:
Wow, that's lots of reviewing you did there :) (so glad btw)

1. about the `array` part you mentioned:
    Yes, I was looking for a way to give access to the original array (and that's all), which recalculating it in `O(n)`(yes it can be calculated in `O(n)` time) is better than storing redundant variables. 

2. Indexing:
    Indexing of a Fenwick-tree is actually on an 1-based array, and there is an extra cell at the beginning. All those `+1` and `-1`s are for shifting array by 1 index. If it affects readability that much (which i think it does!) i'll rewrite it as it needs to be.


3. `index(minimumSum:)`: yes you're right, it should be `Int?`

All fix these and push another commit.  👍🏼


btw, thank you very much for reviewing it, it was my first time writing a readme from scratch (and I think it was obvious at some point too :D)

kelvinlauKL:
@farzadshbfn Just checking in since it's been a while.

farzadshbfn:
@kelvinlauKL To be honest I totally forgot about this. I was super busy with work last couple of months. I'll try to finish it in coming two weeks.

kelvinlauKL:
@farzadshbfn Sounds good, just checking. Thanks!

kelvinlauKL:
Closing this PR due to inactivity. Feel free to open if you still plan on finishing this up!

PR #257, opened 2016-10-14
Fix method name from reverse() to reversed() in README.markdown

keitaito:
Hi! 
I found `reverse()` method was not updated to `reversed()` in README.markdown, so I updated it.
Thanks :)


chris-pilcher:
Thanks @keitaito 👍

keitaito:
@chris-pilcher No problem and thank you for merging! 😄

PR #258, opened 2016-10-15
Use integers only in Queue-Optimized.swift

chrisamanse:


hollance:
Personally, I don't find this change improves the readability of the code. It's a small performance optimization but one that probably doesn't matter in practice. 

If this was code that was executed thousands of times per second then sacrificing the readability for performance may be worth it, but since the goal of this project is to educate people about algorithms and data structures, readability should come first.

(That said, it's a good trick. :-) Maybe we could have an "optimization tips and tricks" section somewhere that includes this particular technique.)

kelvinlauKL:
I agree with @hollance. Having the percentage calculation gives a clear indicator on why the array needs to be resized.

Nice optimization trick though @chrisamanse!

PR #261, opened 2016-10-18
Added public init and array init

RedQueenCoder:
I added a public init so this can be used in a Playground. I also added
an extension to create a linked list from an array of values.


kelvinlauKL:
Thanks @RedQueenCoder !

PR #259, opened 2016-10-15
Migrated Topological Sort to Swift 3

rick527009:
#181. This is my first contribution to this repo. Suggestions are welcome. :)

PR #264, opened 2016-10-21
Migrate topological search swift3

JaapWijnen:
Migrated Topological search to swift 3 for #181


chris-pilcher:
@JaapWijnen It looks like this algorithm has already been updated to Swift 3 by #259.

Sorry! We're trying to close PRs as quickly as possible so this doesn't happen.

PR #182, opened 2016-08-18
Migrating two-sum problem to Swift 3

Jxrgxn:
Can have this one up in a jiffy.

PR #266, opened 2016-10-21
Two sum migrate swift3

JaapWijnen:
Migrated the two sum problem to swift3 as per #181 and due to #182 being closed due to inactivity.


chris-pilcher:
Thanks @JaapWijnen 👍

PR #268, opened 2016-10-24
Simplify midpoint calculation

chrisamanse:
Simplify midpoint calculation by using `(x1 + x2) / 2`. This replaces `x1 + (x2 - x1) / 2`.


chrisamanse:
I just read the note about cases of integer overflow. That's a nice catch. It might be good to place it right below the `midIndex` calculation.

PR #269, opened 2016-10-25
Strassen's Matrix Multiplication Algorithm

rwash8347:
Claim of Strassen's Matrix Multiplication Algorithm. Any suggestions/comments let me know! 😄


Strassen Matrix Multiplication/Matrix.swift
```diff
@@ -0,0 +1,246 @@
...
+//  Matrix.swift
+//  
+//
+//  Created by Richard Ash on 10/28/16.
+//
+//
+
+import Foundation
+
+enum RowOrColumn {
```
kelvinlauKL: Since you're only using this internally for the `subscript`, it feels a bit better to put it in the `Matrix` namespace.

Strassen Matrix Multiplication/Matrix.swift
```diff
@@ -0,0 +1,246 @@
...
+      case .column:
+        assert(newValue.count == rows)
+        for (row, element) in newValue.enumerated() {
+          grid[(row * columns) + value] = element
+        }
+      }
+    }
+  }
+  
+  // MARK: - Private Functions
```
kelvinlauKL: Consider putting all the `private` methods in a `private` extension. It's pretty nice in that it visually groups the helper methods together, and has a few architectural benefits as well - nobody would be able to put a differently scoped method / computed property in the middle of your `private` section.

Strassen Matrix Multiplication/Matrix.swift
```diff
@@ -0,0 +1,246 @@
...
+    }
+    
+    return C
+  }
+  
+  private func nextPowerOfTwo(of n: Int) -> Int {
+    return Int(pow(2, ceil(log2(Double(n)))))
+  }
+  
+  // MARK: - Functions
```
kelvinlauKL: Similarly to the `private` extension discussion above, I suggest putting the public / internal stuff in an extension.

Strassen Matrix Multiplication/MatrixSize.swift
```diff
@@ -0,0 +1,26 @@
...
+  func forEach(_ body: (Int, Int) throws -> Void) rethrows {
+    for row in 0..<rows {
+      for column in 0..<columns {
+        try? body(row, column)
+      }
+    }
+  }
+}
+
+func ==(lhs: MatrixSize, rhs: MatrixSize) -> Bool {
```
kelvinlauKL: As of Swift 3.0, you can nest the `Equatable` method requirement inside the conforming type's scope:

```
extension MatrixSize: Equatable {
  static func ==(lhs: MatrixSize, rhs: MatrixSize) -> Bool {
    return lhs.columns == rhs.columns && lhs.rows == rhs.rows
  }
}
```

You just need to prepend the `static` modifier.

Strassen Matrix Multiplication/README.markdown
```diff
@@ -0,0 +1 @@
+## TODO: Examples & Explanation
```
kelvinlauKL: Looking forward to this! Linear Algebra seems to be a prerequisite for this article, so it'll be good to mention they should have some background on working with Matrices, and provide a link to where you would recommend them to get some basics down before looking into this.

You're also making use of your own helper protocols and defining your own overloaded operators, so make sure you make clear references to avoid confusion for the readers when they follow along.

Strassen Matrix Multiplication/Matrix.swift
rwash8347: Yea definitely. I tried this initially but kept getting the error "Type 'RowOrColumn' cannot be nested in generic type 'Matrix'". Do you know of a workaround?

kelvinlauKL: Opps, you're right. I believe this was a discussion point as part of the generics manifesto that's planned for Swift 4.0.

I don't think a workaround is possible at the moment.

rwash8347:
Thanks for the comments! Just updated the readme, take a look and let me know if you have any suggestions!

Strassen Matrix Multiplication/StrassensMatrixMultiplication.playground/Sources/Matrix.swift
```diff
@@ -0,0 +1,283 @@
...
+//  Matrix.swift
+//  
+//
+//  Created by Richard Ash on 10/28/16.
+//
+//
+
+import Foundation
+
+public enum RowOrColumn {
```
kelvinlauKL: As of Swift 3.1, this can now be nested within generic types. 🎉

rwash8347: I know! So excited 💯🎉 , I'll make the change!

Strassen Matrix Multiplication/README.markdown
```diff
@@ -0,0 +1,468 @@
...
+}
+
+for i in B.rows {
+  for j in B.columns {
+    BPrep[i, j] = B[i, j]
+  }
+}
+```
+
+Finally, we recusrisvely compute the matrix using Strassen's algorithm and the transform our new matrix `C` back to the correct dimensions!
```
kelvinlauKL: Prefer `recursively` rather than `recusrisevly` 👍

rwash8347: 🤦‍♂️ haha can't believe I missed that

rwash8347:
Awesome! Thanks again for the comments @kelvinlauKL. Went through the readme again and fixed all the typos!

PR #271, opened 2016-10-27
migrate fixedsizearray to swift3

JaapWijnen:
#181


kelvinlauKL:
I've made a few changes on this:

To keep consistent with the other native `Collection` types, I've removed the argument labels for `append`.

`removeAtIndex(index: Int)` has also been changed to `removeAt(index: Int)` to avoid mentioning index twice.

Thanks!

JaapWijnen:
Ah good point! I'll try to think of it next time.

PR #277, opened 2016-11-01
took off an extra min

BethKnight1234:
I took off an extra min in the your random function. This is my first open source contribution!! Ever!


kelvinlauKL:
Hey @BethKnight1234

Thanks for your enthusiasm! Good catch, I'm merging this in.

PR #279, opened 2016-11-02
Fix for a bug in SegmentTree

bobawhaler:


kelvinlauKL:
Thanks for the catch @bobawhaler!

PR #280, opened 2016-11-05
Fix memory leak in Tree implementation

chrisamanse:
Caused by retain cycle between parent and child (parent has a strong reference to child, and child has a strong reference to parent).

Related to #92.


kelvinlauKL:
Cheers.

PR #293, opened 2016-11-17
Covert some codes to Swift3

DingHub:


DingHub:
Covert some codes to Swift3

PR #294, opened 2016-11-17
Made Peek() a computed property

codeOfRobin:
IMHO, making peek a computed property feels _swiftier_

### DIFF
--- a/Queue/Queue.playground/Contents.swift
+++ b/Queue/Queue.playground/Contents.swift
@@ -34,7 +34,7 @@ public struct Queue<T> {
     }
   }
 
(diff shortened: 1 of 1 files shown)


kelvinlauKL:
Yep, I agree @codeOfRobin. I've been especially conscious lately of making any function that takes no parameters as a computed property instead. Good catch!

hollance:
I have a small comment on this. ;-) "peek" is a verb, which makes sense for a function. Properties should be nouns, not verbs. So if you're making this a computed property, then you might want to name it `front` or something like that.

PR #310, opened 2016-12-06
Fixes removal method bug for Trie

kelvinlauKL:
@Zampano pointed out this error in the removal algorithm -[https://forums.raywenderlich.com/t/swift-algorithm-club-swift-trie-data-structure/21509/2](link).

This PR updates a few things:

1. The `remove` method now correctly removes a word that isn't part of another word.
2. Added a `words` property of type `Set<String>` to facilitate <em>O(1)</em> querying of the `count`, `isEmpty`, and `word` property of the Trie.


rzaccone:
Having a <code>words</code> property that is a set raises some issues.  It seems that the trie becomes unnecessary because operations could instead be performed on the set.  For example, a really simple implementation of <code>contains</code> would just check to see if the set contained a word.  A set property would probably more than double the amount of storage needed for the trie too.  It would be nice to have <code>words</code> be an O(1) operation, but I think it should be a traversal.

kelvinlauKL:
@rzaccone Good point. Is it expected for a Trie to have a <em>O(1)</em> looking for the list of words inside it anyway?

Perhaps we should just change `words` into a computed property traverses through all the nodes to aggregate all the words, making it a <em>O(n)</em> operation.

rzaccone:
I searched through some of my books, including Knuth's Sorting and Searching and couldn't find this issue addressed.  I think that having this be a property that is computed by traversing the nodes is the way to go.  This would be an O(n) operation where n is the number of nodes in the trie.

rzaccone:
I've done some more refactoring.  I also added the <code>words</code> property, <code>isEmpty</code> and <code>count</code>.  I tried to test the trie using a large file of words (over 162,000).  The playground crashes and it's very slow!  I moved everything over to a Cocoa project and wrote some XCTests.  Everything seems to be working and the performance seems fairly good.  What's the best way to share what I've done?

kelvinlauKL:
Cool!

Fork the repository, and make a pull request with the changes you've made. Then we'll do some peer review to make sure everything works as intended. 

Don't forget to also update the `README.markdown` file to reflect the new changes.

kelvinlauKL:
Closed due to a fix by #313

jargomarg:
Loved the guide!

I was trying to create a recursive function that could search the Trie for words contained in an anagram. For example, it might take a scrambled string "rehnla" and would return and array of strings ["learn","heal","hen", etc.]. The idea was described here http://stackoverflow.com/a/880611

Any ideas on how I could try and implement this?

PR #313, opened 2016-12-12
I created an Xcode project with my changes.  It makes it easier to

rzaccone:
I created an Xcode project with my changes.  It makes it easier to test and the playground was crashing!


Trie/Trie/Trie/Trie.swift
```diff
@@ -0,0 +1,207 @@
...
+    ///
+    /// - Parameter word: the word to be inserted.
+    func insert(word: String) {
+        guard !word.isEmpty else {
+            return
+        }
+        var currentNode = root
+        let charactersInWord = Array(word.lowercased().characters)
+        var index = 0
+        while index < charactersInWord.count {
```
kelvinlauKL: A `for` loop can do exactly what we're trying to do in a cleaner manner in this scenario (I know the site tutorial uses the `while` construct, but a fellow contributor alerted me of a better way).

```
for character in word.lowercased().characters {
  if let childNode = currentNode.children[character] {
    currentNode = childNode
  } else {
    currentNode.add(value: character)
    currentNode = currentNode.children[character]!
  }
}
```

This removes the necessity of the `index` and `charactersInWord` properties. The `for` construct also helps us bind each element in the `word.lowercased().characters`, alleviating the necessity of picking it out ourselves.

Trie/Trie/Trie/Trie.swift
```diff
@@ -0,0 +1,207 @@
...
+                currentNode.add(value: character)
+                currentNode = currentNode.children[character]!
+            }
+            index += 1
+        }
+        // Word already present?
+        guard !currentNode.isTerminating else {
+            return
+        }
+        self.wordCount += 1
```
kelvinlauKL: Unnecessary reference to `self`.

Trie/Trie/Trie/Trie.swift
```diff
@@ -0,0 +1,207 @@
...
+        currentNode.isTerminating = true
+    }
+
+    /// Determines whether a word is in the trie.
+    ///
+    /// - Parameter word: the word to check for
+    /// - Returns: true if the word is present, false otherwise.
+    func contains(word: String) -> Bool {
```
kelvinlauKL: Replacing this with a `for` loop would simplify things here too:

```
func contains(word: String) -> Bool {
  guard !word.isEmpty else {
    return false
  }
    
  var currentNode = root
  for character in word.lowercased().characters {
    guard let childNode = currentNode.children[character] else { return false }
    currentNode = childNode
  }
    
  return currentNode.isTerminating
}
```

Trie/Trie/Trie/Trie.swift
rzaccone: I agree.  This is a good change.

Trie/Trie/Trie/Trie.swift
rzaccone: Again, I agree.

rzaccone:
Here's a slight simplification to the `for` loop in `wordsInSubtrie`.
```
        for childNode in rootNode.children.values {
            let childWords = wordsInSubtrie(rootNode: childNode, partialWord: previousLetters)
            subtrieWords += childWords
        }
```

rzaccone:
Along the lines of your other good suggestions, here's an improvement to `findTerminalNodeOf`.  It's the same adjustment to the `for` loop.
```
    private func findTerminalNodeOf(word: String) -> Node? {
        var currentNode = root
        for character in word.lowercased().characters {
            guard let childNode = currentNode.children[character] else {
                return nil
            }
            currentNode = childNode
        }
        return currentNode.isTerminating ? currentNode : nil
    }
```

kelvinlauKL:
Yep, things are looking pretty Swifty :)

Could you update the PR with the changes?

rzaccone:
I pushed all my changes and all tests pass.

kelvinlauKL:
Looks great. I took a look over everything, and all tests passed. 

I'll merge this in now. Cheers!

PR #315, opened 2016-12-14
Update enum case style in Binary Tree

pixyzehn:
In Binary Tree, updated enum case style.
Also, updated README in Binary Tree.


kelvinlauKL:
Cheers!

PR #330, opened 2016-12-24
Boyer-Moore algorithm updates

### LINKED ISSUE #284: Duplicate code in Boyer-Moore String Search.md
In [Boyer-Moore String Search.md](https://github.com/raywenderlich/swift-algorithm-club/tree/master/Boyer-Moore), the code of Boyer-Moore-Horspool algorithm is the same as that of Boyer-Moore algorithm. It doesn't show the difference in how to check partial matches.

mmazzei:
This is what I did:
 - added some unit tests
 - ensure every test pass
 - added Boyer-Moore-Horspool algorithm (previously mentioned in the README but don't implemented) with tests.
 - added me in the credits section ;)

This fixes #284.

### DIFF
--- a/Boyer-Moore/BoyerMoore.playground/Contents.swift
+++ b/Boyer-Moore/BoyerMoore.playground/Contents.swift
@@ -1,72 +1,91 @@
 //: Playground - noun: a place where people can play
 
+/*
+  Boyer-Moore string search
+
+  This code is based on the article "Faster String Searches" by Costas Menico
+  from Dr Dobb's magazine, July 1989.
+  http://www.drdobbs.com/database/faster-string-searches/184408171
+*/
 extension String {
-  func indexOf(pattern: String) -> String.Index? {
-    // Cache the length of the search pattern because we're going to
-    // use it a few times and it's expensive to calculate.
-    let patternLength = pattern.characters.count
-    assert(patternLength > 0)
-    assert(patternLength <= characters.count)
-    
-    // Make the skip table. This table determines how far we skip ahead
-    // when a character from the pattern is found.
-    var skipTable = [Character: Int]()
-    for (i, c) in pattern.characters.enumerated() {
-      skipTable[c] = patternLength - i - 1
+    func index(of pattern: String, usingHorspoolImprovement: Bool = false) -> Index? {
+        // There are no possible match in an empty string
+        guard !isEmpty else { return nil }
+
+        // Cache the length of the search pattern because we're going to
+        // use it a few times and it's expensive to calculate.
+        let patternLength = pattern.characters.count
+        guard patternLength > 0, patternLength <= characters.count else { return nil }
+
+        // Make the skip table. This table determines how far we skip ahead
+        // when a character from the pattern is found.
+        var skipTable = [Character: Int]()
+        for (i, c) in pattern.characters.enumerated() {
+            skipTable[c] = patternLength - i - 1
+        }
+
+        // This points at the last character in the pattern.
+        let p = pattern.index(before: pattern.endIndex)
+        let lastChar = pattern[p]
+
+        // The pattern is scanned right-to-left, so skip ahead in the string by
+        // the length of the pattern. (Minus 1 because startIndex already points
+        // at the first character in the source string.)
+        var i = index(startIndex, offsetBy: patternLength - 1)
+
+        // This is a helper function that steps backwards through both strings
+        // until we find a character that doesn’t match, or until we’ve reached
+        // the beginning of the pattern.
+        func backwards() -> Index? {
+            var q = p
+            var j = i
+            while q > pattern.startIndex {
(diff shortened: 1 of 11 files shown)


Boyer-Moore/BoyerMoore.swift
```diff
@@ -6,7 +6,7 @@
   http://www.drdobbs.com/database/faster-string-searches/184408171
 */
 extension String {
-    func indexOf(pattern: String) -> String.Index? {
+    func indexOf(pattern: String, useHorspoolImprovement: Bool = false) -> String.Index? {
```
kelvinlauKL: I think `index(of pattern: String, useHorspoolImprovement: Bool = false) -> Index?` is a better method signature. 

Since Swift 3, many methods switched over from `indexOf` to `index(of:)`. `String.Index` is unnecessary since we're inside a `String` extension and the compiler can infer the type of `Index`.

Boyer-Moore/BoyerMoore.swift
```diff
@@ -53,15 +53,23 @@ extension String {
                 // There is a possible match. Do a brute-force search backwards.
                 if let k = backwards() { return k }
                 
-                // If no match, we can only safely skip one character ahead.
-                i = index(after: i)
+                if !useHorspoolImprovement {
+                    // If no match, we can only safely skip one character ahead.
+                    i = index(after: i)
+                }
+                else {
```
kelvinlauKL: Our style guidelines prefer to have the `else` statement on the same line as the closing brackets for the `if` statement.

Boyer-Moore/README.markdown
```diff
@@ -158,40 +158,66 @@ Here's an implementation of the Boyer-Moore-Horspool algorithm:
 ```swift
 extension String {
     func indexOf(pattern: String) -> String.Index? {
-    let patternLength = pattern.characters.count
-    assert(patternLength > 0)
-    assert(patternLength <= self.characters.count)
+        // Cache the length of the search pattern because we're going to
+        // use it a few times and it's expensive to calculate.
+        let patternLength = pattern.characters.count
+        assert(patternLength > 0)
```
kelvinlauKL: I feel like we should replace these asserts with `guard patternLength > 0, characters.count < patternLength else { return nil }`. What do you think?

Boyer-Moore/README.markdown
```diff
@@ -158,40 +158,66 @@ Here's an implementation of the Boyer-Moore-Horspool algorithm:
...
-    i = index(after: i)
-    } else {
-    i = index(i, offsetBy: skipTable[c] ?? patternLength)
-    }
-    }
-    return nil
+        // The pattern is scanned right-to-left, so skip ahead in the string by
+        // the length of the pattern. (Minus 1 because startIndex already points
+        // at the first character in the source string.)
+        var i = self.index(startIndex, offsetBy: patternLength - 1)
```
kelvinlauKL: Unnecessary reference to `self`. There are a couple of these.

mmazzei:
Thanks for the comments!

I agree with the improvements and implemented the changes.

Also updated the playground and made other minor changes (added tests, renamed `useHorspoolImprovement` to `usingHorspoolImprovement`).

The new Swift 3 style for method signatures feels more natural.

Regards!

Boyer-Moore/BoyerMoore.playground/Contents.swift
```diff
@@ -1,72 +1,91 @@
...
-        // The characters are not equal, so skip ahead. The amount to skip is
-        // determined by the skip table. If the character is not present in the
-        // pattern, we can skip ahead by the full pattern length. However, if
-        // the character *is* present in the pattern, there may be a match up
-        // ahead and we can't skip as far.
-        i = index(i, offsetBy: skipTable[c] ?? patternLength)
-      }
+    func index(of pattern: String, usingHorspoolImprovement: Bool = false) -> Index? {
+        // There are no possible match in an empty string
+        guard !isEmpty else { return nil }
```
kelvinlauKL: This feels a bit redundant since the latter `guard` statement covers this case as well.

Boyer-Moore/BoyerMoore.swift
```diff
@@ -6,33 +6,35 @@
   http://www.drdobbs.com/database/faster-string-searches/184408171
 */
 extension String {
-    func indexOf(pattern: String) -> String.Index? {
+    func index(of pattern: String, usingHorspoolImprovement: Bool = false) -> Index? {
+        // There are no possible match in an empty string
+        guard !isEmpty else { return nil }
```
kelvinlauKL: Same as before, the latter `guard` statement should cover this case just as well.

mmazzei:
You're right. Redundant validation removed. 👍

kelvinlauKL:
Merging this in! Thanks and happy Boxing Day! @mmazzei

PR #335, opened 2016-12-29
Updating to Swift3 (enum lowercase)

cruisediary:
update enum case uppercase to lowercase 
- playground
- README.md
- some.swift file


vincentngo:
@cruisediary looks good thanks!

PR #336, opened 2016-12-29
Serialize with array

rzaccone:
This is my attempt to serialize the trie.  I invite suggestions on how to improve this.  As I explain in the ReadMe file.  There are issues with this approach.


Trie/Trie/Trie/ReadMe.md
```diff
@@ -0,0 +1,33 @@
...
+
+There are several downsides though.
+
+- The size of the archived words is three times the size of the original file of words!  Did I do this right?  The tests pass, so it seems correct.  I question whether archiving is worth the effort if the resulting archive is so much larger than the original file.
+
+- I would expect that archiving the trie would result in a file that was smaller than the original since a trie doesn't repeat leading character sequences when they are the same.
+
+- This requires that the trie get reconstructed when it is unarchived.
+
+- My gut tells me that it would be faster to archive and unarchive the trie itself, but I don't have any hard data to support this.
```
kelvinlauKL: Will add this to the wish list in case anyone knows and is willing to contribute :)

Trie/Trie/Trie/Trie.swift
```diff
@@ -62,11 +60,32 @@ class Trie {
     fileprivate let root: Node
     fileprivate var wordCount: Int
 
-    /// Creats an empty trie.
-    init() {
+    /// Creates an empty trie.
+    override init() {
```
kelvinlauKL: Why didn't you call the `super` init for `NSObject` here?

Trie/Trie/TrieTests/TrieTests.swift
```diff
@@ -48,24 +48,23 @@ class TrieTests: XCTestCase {
...
+
     /// Inserts words into a trie.  If the trie is non-empty, don't do anything.
     func insertWordsIntoTrie() {
         guard trie.count == 0 else {
             return
         }
-        let numberOfWordsToInsert = wordArray!.count
-        for i in 0..<numberOfWordsToInsert {
-            trie.insert(word: wordArray![i])
+        for word in self.wordArray! {
```
kelvinlauKL: I would unwrap `wordArray` in the `guard` statement above:

```swift
guard let wordArray = wordArray, trie.count == 0 else { return }

for word in wordArray {
  // insert
}
```

Trie/Trie/Trie/Trie.swift
rzaccone: I didn't know I needed to do that.  Doesn't the compiler complain if you need to call super.init()?

Trie/Trie/TrieTests/TrieTests.swift
rzaccone: This seems like a good suggestion.

Trie/Trie/Trie/Trie.swift
kelvinlauKL: You bring up a good point, I'm not sure why the compiler didn't complain in this situation, but according to the `NSObject` [docs](https://developer.apple.com/reference/objectivec/nsobject/1418641-init): 

> In a custom implementation of this method, you must invoke super’s Initialization then initialize and return the new object.

Personally speaking, I always call `super` whenever I override anything. Haven't come across a situation where not calling `super` provided any benefits. Let me know if you have another perspective on this.

kelvinlauKL: I asked around and came up with this conclusion - When subclassing, always call the `super` implementation. 

In this case, the compiler did not complain because of a very special circumstance where the superclass (`NSObject`) had only 1 initializer. In this case, Swift will automatically call the superclasses initializer for you. 

According to the person I was speaking to, there has been some compiler bugs in the past where inserting the explicit `super.init()` actually changes behaviour. 

In light of this, I think doing an explicit `super.init()` is the right way to go here.

rzaccone: I agree that we shouldn't rely on some special compiler behavior.  I find this type of special behavior annoying.  I can't remember special cases like this!  It's better to do it explicitly.

I made both changes and pushed them.  If there's something else I need to do, let me know.  I sometimes find git overwhelming!

kelvinlauKL:
Looks good. Before I merge, there's a merge conflict with your fork and the master branch of this repo. Could you fix that?

rzaccone:
Resolved the merge conflict.

kelvinlauKL:
@rzaccone Was this closed by mistake?

rzaccone:
I must have pressed the wrong button.  Sorry.

kelvinlauKL:
Thanks @rzaccone!

PR #337, opened 2016-12-30
Added a new algorithms book under "Learn More"

yoha:
This book is different from other algorithms books for its rich illustrations. It's immensely helpful for those who are visual learner like myself.


kelvinlauKL:
Sounds good. Thanks for the recommendation 👍.

Happy New Years!

PR #338, opened 2016-12-30
Adds deinit to LinkedList

wokalski:
Fixes potential retain cycle if the linked list has a circular reference.

### DIFF
--- a/Linked List/LinkedList.swift
+++ b/Linked List/LinkedList.swift
@@ -137,6 +137,14 @@ public class LinkedList<T> {
     assert(node != nil)
     return remove(node: node!)
   }
+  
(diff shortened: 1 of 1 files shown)


vincentngo:
@wokalski Hm i don't think it would be possible to cause a retain cycle if the linked list is circular. Since the previous property is already defined as weak.

wokalski:
It would be possible because `next` is strong. So if you do `node.next.next = node` you created a cycle. Not applicable here though because there's no access to next from outside the project if I understand correctly. Apologies.

vincentngo:
@wokalski sounds good, thanks. Good insight though!

PR #343, opened 2017-01-02
Updating Shortest Path (Unweighted) to Swift 3

gonini:
I upgraded to swift 3 because it did not run as a grammar problem.


kelvinlauKL:
Looks good, thanks @gonini!

PR #344, opened 2017-01-03
Migrated Huffman Coding to Swift 3 syntax

pbodsk:
Hope I got this right 😄

PR #346, opened 2017-01-05
Migrates All-Pairs Shortest Paths to Swift 3 syntax

pbodsk:


pbodsk:
Sorry...got two pull requests squashed into one. Huffman Coding is already in #344 😊

PR #352, opened 2017-01-09
Migrates Ternary Search Tree to Swift 3 syntax

pbodsk:


Ternary Search Tree/TST.playground/Contents.swift
```diff
@@ -25,18 +25,18 @@ var testStrings: [(key: String, data: String)] = []
...
-    let key = randomAlphaNumericString(randomLength)
-    let data = randomAlphaNumericString(randomLength)
+    let key = randomAlphaNumericString(withLength: randomLength)
+    let data = randomAlphaNumericString(withLength: randomLength)
 //    print("Key: \(key) Data: \(data)")
 
     if key != "" && data != "" {
         testStrings.append((key, data))
-        treeOfStrings.insert(data, withKey: key)
+        treeOfStrings.insertData(data, withKey: key)
```
vincentngo: Would be cool if this function parameter would read:

`treeOfStrings.insert(data: data, withKey: key)`

Ternary Search Tree/TST.playground/Sources/TernarySearchTree.swift
```diff
@@ -17,12 +17,12 @@ public class TernarySearchTree<Element> {
 
     //MARK: - Insertion
 
-    public func insert(data: Element, withKey key: String) -> Bool {
+    public func insertData(_ data: Element, withKey key: String) -> Bool {
```
vincentngo: Do you think the function would look cleaner if we did the following?
func insert(data: Element, withKey key: String) -> Bool  {}

Ternary Search Tree/TST.playground/Contents.swift
vincentngo: following the swift api guidelines here: https://swift.org/documentation/api-design-guidelines/#naming

pbodsk:
You're absolutely right @vincentngo  

The reason I kept `insertData` instead of renaming it to `insert(data:)` was because there are some private methods called `findNode` and `insertNode` in the `TernarySearchTree` class. 

But...I agree, `insert(data: data, withKey: key)` looks better, so I'll change that. And to align things I'll rename the private functions `findNode` and `insertNode` so the follow the naming conventions too.

Stay tuned 😄

pbodsk:
Aaaand done. 

I took the liberty of adding some unit tests as well (just a couple) and they seem to pass 😌 

Hope it is as expected

Ternary Search Tree/TernarySearchTree.swift
```diff
@@ -95,40 +95,40 @@ public class TernarySearchTree<Element> {
...
      - parameter charIndex: The index of the current char in the search key
 
      - returns: The element, if found. Nil otherwise.
      */
-    private func findNode(aNode: TSTNode<Element>?, withKey key: String, atIndex charIndex: Int) -> Element? {
+    private func find(node: TSTNode<Element>?, withKey key: String, atIndex charIndex: Int) -> Element? {
 
         //Given key does not exist in tree.
-        if aNode == nil {
+        if node == nil {
```
vincentngo: I think we could convert this to a guard statement and return nil if node doesn't exist.

Ternary Search Tree/TernarySearchTree.swift
```diff
@@ -95,40 +95,40 @@ public class TernarySearchTree<Element> {
...
             return nil
         }
 
-        let index = key.startIndex.advancedBy(charIndex)
+        let index = key.index(key.startIndex, offsetBy: charIndex)
         //go left
-        if key[index] < aNode!.key {
-            return findNode(aNode!.left, withKey: key, atIndex: charIndex)
+        if key[index] < node!.key {
```
vincentngo: So we don't have to force unwrap node 👍

Ternary Search Tree/TernarySearchTree.swift
```diff
@@ -33,52 +33,52 @@ public class TernarySearchTree<Element> {
...
         }
 
         //create a new node if necessary.
-        if aNode == nil {
-            let index = key.startIndex.advancedBy(charIndex)
-            aNode = TSTNode<Element>(key: key[index])
+        if node == nil {
+            let index = key.index(key.startIndex, offsetBy: charIndex)
+            node = TSTNode<Element>(key: key[index])
         }
```
vincentngo: You can probably add a guard statement here, and return false for whatever reason the node failed to create.

Ternary Search Tree/TernarySearchTree.swift
```diff
@@ -33,52 +33,52 @@ public class TernarySearchTree<Element> {
...
 
         //if current char is less than the current node's char, go left
-        let index = key.startIndex.advancedBy(charIndex)
-        if key[index] < aNode!.key {
-            return insertNode(&aNode!.left, withData: data, andKey: key, atIndex: charIndex)
+        let index = key.index(key.startIndex, offsetBy: charIndex)
+        if key[index] < node!.key {
```
vincentngo: then you can just do node.key instead of node!.key

vincentngo:
Also it would be awesome if you had time to add some README doc on how ternary search trees work! But that's not part of this pull request haha.

Ternary Search Tree/TernarySearchTree.swift
pbodsk: 👍 done

Ternary Search Tree/TernarySearchTree.swift
pbodsk: About this one....I'm having doubts about how to fix this. It seems that central parts of the algorithm is based on creating new nodes if the passed in node is `nil`.

The nodes used are of type `TSTNode`, and they contain three children: `left`, `right` and `middle` and these are then used when we insert a new node in the tree.

The first time you call this method, you do so with the `root` node. That is `nil` at that time, so you create a new value and use that.

Later on, when building your tree, you look at the key to determine whether you should do a recursive call to `insert` with the left, middle or right child node. At first the children are `nil` too, so they are created when `insert` is called.

So...you can probably get around the `nil` check in some way, but I cannot figure out how right now. 

Input is welcome 😄

vincentngo: Fair enough haha. I think this is pretty good for now! Thanks again.

PR #355, opened 2017-01-11
fixing failing test on Ternary Search Tree

pbodsk:
The problem was that the testdata generator could sometimes (but not on every test run...) generate keys that was not unique...which caused the `find` method to return the wrong values for keys when more keys existed in the tree with the same value.

### DIFF
--- a/Ternary Search Tree/Tests/TernarySearchTreeTests.swift
+++ b/Ternary Search Tree/Tests/TernarySearchTreeTests.swift
@@ -10,14 +10,23 @@ import XCTest
 
 class TernarySearchTreeTests: XCTestCase {
     
-    let treeOfStrings = TernarySearchTree<String>()
+
     let testCount = 30
     
     func testCanFindStringInTree() {
         var testStrings: [(key: String, data: String)] = []
+        let treeOfStrings = TernarySearchTree<String>()
+
         for _ in (1...testCount) {
-            let randomLength = Int(arc4random_uniform(10))
-            let key = Utils.shared.randomAlphaNumericString(withLength: randomLength)
+            var randomLength = Int(arc4random_uniform(10))
+            
+            var key = Utils.shared.randomAlphaNumericString(withLength: randomLength)
+            
+            while testStrings.contains(where: { $0.key == key}) {
+                //That key is taken, so we generate a new one with another length
+                randomLength = Int(arc4random_uniform(10))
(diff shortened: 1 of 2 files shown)


vincentngo:
I was wondering why the test were sometimes passing and sometimes failing randomly :D

Ternary Search Tree/Tests/TernarySearchTreeTests.swift
```diff
@@ -10,14 +10,23 @@ import XCTest
...
     
-    let treeOfStrings = TernarySearchTree<String>()
+
     let testCount = 30
     
     func testCanFindStringInTree() {
         var testStrings: [(key: String, data: String)] = []
+        let treeOfStrings = TernarySearchTree<String>()
+
         for _ in (1...testCount) {
```
vincentngo: @pbodsk would it be possible to build the TST once and use that tree to perform your test? Instead of creating it on every test?

vincentngo: You could extract **testStrings** and **treeOfStrings** into properties of XCTestCase, and perform a setup to generate the tree. That way each test focuses on what it's testing.

pbodsk: I see you point but...

The first test: `testCanFindStringInTree`, generates a tree containing `String` elements so therefore the `testStrings` and `treeOfStrings` are created there, like so:

        var testStrings: [(key: String, data: String)] = []
        let treeOfStrings = TernarySearchTree<String>()

The second test: `testCanFindNumberInTree`, generates a tree containing `Int` elements, so here the variables used looks like this:

        var testNums: [(key: String, data: Int)] = []
        let treeOfInts = TernarySearchTree<Int>()

(notice that one tree works on `String` elements and the other works on `Int` elements).

If all tests were using the same tree structure then I agree that it would be way better to generate/populate the `treeOfStrings` and the `testStrings` in `setup()`, but now the two tests are using different trees for their testing, and I think adding them as two different properties on the testcase would clutter the testclass.

If we add more tests at a later point then perhaps we should consider extracting `treeOfStrings` into a property and then do tests only with `String` elements.

Hope this makes sense to you (not trying to dodge the task here 😄)

vincentngo: Ah ok fair enough XD

PR #356, opened 2017-01-13
migrates Single-Source Shortest Paths to Swift 3 syntax

pbodsk:


pbodsk:
hi @vincentngo 

I would love to do so....but I'm terribly sorry, I'm not sure I understand what you mean. The `SSSP Test Target` has been converted to Swift 3 as well.

Where do I toggle the legacy code setting?

Is it this?

screen shot 2017-01-14 at 09 28 06

vincentngo:
@pbodsk yup that's right. The reason why is because when someone opens the xcworkspace, it will show a popup saying " Would you like to convert to latest Swift", but you already did it! :)

pbodsk:
@vincentngo OK...and should that be set to `no`?

vincentngo:
@pbodsk yup

vincentngo:
@pbodsk i'm going to go ahead and merge this, and set the flag for you 👍  since it's trivial

pbodsk:
Thanks @vincentngo

PR #360, opened 2017-01-19
Simplify the insert method

boqian2000:
I made two changes to the Binary Search Tree:
1.  Simplify the insert method by removing the fileprivate method.  
2. Delete duplicated BinarySearchTree.swift, so that there is only one single file.


kelvinlauKL:
Looks good. Thanks @boqian2000!

PR #361, opened 2017-01-20
Simplified the remove()

boqian2000:
Simplified the remove().  I feel the new implementation is more elegant (cutting lines of code from 32 to 18) and conceptually more straightforward.


Binary Search Tree/Solution 1/BinarySearchTree.playground/Sources/BinarySearchTree.swift
```diff
@@ -350,24 +322,23 @@ extension BinarySearchTree: CustomStringConvertible {
...
+   public func toArray() -> [T] {
+      return map { $0 }
+   }
 
-  public func toArray() -> [T] {
-    return map { $0 }
-  }
 }
+
+//extension BinarySearchTree: CustomDebugStringConvertible {
```
boqian2000: That was a mistake. I didn't intend to check it in.  However, I do feel the description is much more readable than the debugDescription.  It makes the playground much easier and fun to play with. So I still would like to recommend removing CustomDebugStringConvertible().

kelvinlauKL:
@boqian2000 Could you update the `README` file with the updated logic too? 

I agree with your deletion on `CustomDebugStringConvertible` - initially I thought you deleted `CustomStringConvertible`. 

Your change feels good, and the previous implementation really felt clunky.

boqian2000:
Done updating the README file.

kelvinlauKL:
Merged. Thanks @boqian2000

PR #366, opened 2017-01-28
Set Cover migrated to swift 3

JaapWijnen:


vincentngo:
@JaapWijnen looks good thanks

PR #382, opened 2017-02-13
Use string interpolation for the whole description string

sigito:


kelvinlauKL:
Good change 👍 Thanks @sigito

Issue #378, opened 2017-02-08
Unexpected behavior in `wordsInSubtrie`


AdrianBinDC:
I was tinkering with this class and I tried to use the `wordsInSubtrie` method to return an array of strings. Here's what I did:

`var trie = Trie()`
`trie.insert(word: "test")`
`trie.insert(word: "another")`
`trie.insert(word: "exam")`
`print(trie.wordsInSubtrie(rootNode: trie.root, partialWord: "exam"))`

I would expect an array with just one element, exam, to be returned in the array. Instead, this array is what is returned:

`["examtest", "examanother", "examexam"]`

kelvinlauKL:
Interesting! Will check that out.

kelvinlauKL:
It seems like `wordsInSubtrie` isn't meant to be used publicly (i.e. it fails to work properly if you supply anything other than an empty string to `partialWord`).

This trie doesn't support prefix matching yet. Sounds like something to add to the Wishlist :). Thanks for the heads up!

kelvinlauKL:
This is fixed!

PR #383, opened 2017-02-13
Trie - Support prefix matching

abhisood:
Solves issue #378 

`findWordsWithPrefix` finds all words starting with a prefix. Includes prefix if it's a word stored in the trei. Returns all words if prefix = "" 

Added unit test.


kelvinlauKL:
Looks good! thanks @abhisood

PR #388, opened 2017-02-22
Adding an implementation for a general-purpose tree using enums

GABRIELFILMSTUDIOS:
Enums are passed by reference, so providing an implementation is useful in situations where value semantics are needed.


Tree/README.markdown
```diff
@@ -151,6 +153,149 @@ tree.search("chai")     // returns the "chai" node
...
+
+sameTree.value   // "drinks"
+tree.value       // also "drinks"
+```
+
+While this can be desired in some cases, in other cases you want to have something, that is passed by value, meaning that it is copied every time you hand it to a function, assign it to a variable, etc. To implement this, we will use and enum:
+
+```swift
+enum TreeNode<T> {
+  case Leaf(value: T)
```
kelvinlauKL: UpperCamelCase for enum cases has been deprecated for lowerCamelCase since Swift 3.

```
enum TreeNode<T> {
  // preferred
  case leaf(value: T)
  indirect case node(value: T, children: [TreeNode])
}
```

Tree/README.markdown
```diff
@@ -151,6 +153,149 @@ tree.search("chai")     // returns the "chai" node
...
+  indirect case Node(value: T, children: [TreeNode])
+  
+  func add(child: TreeNode) -> TreeNode {
+    switch self {
+    case .Leaf(let ownValue):
+      return .Node(value: ownValue, children: [child])
+    case .Node(let ownValue, let ownChildren):
```
kelvinlauKL: You can factor out the `let` prefix on the associated values.

```
case let .node(ownValue, ownChildren)
```

Tree/README.markdown
```diff
@@ -151,6 +153,149 @@ tree.search("chai")     // returns the "chai" node
...
+var greenTeaNode = TreeNode.Leaf(value: "green")
+var chaiTeaNode = TreeNode.Leaf(value: "chai")
+
+var sodaNode = TreeNode.Leaf(value: "soda")
+var milkNode = TreeNode.Leaf(value: "milk")
+
+var gingerAleNode = TreeNode.Leaf(value: "ginger ale")
+var bitterLemonNode = TreeNode.Leaf(value: "bitter lemon")
+
+sodaNode = sodaNode.add(gingerAleNode)
```
kelvinlauKL: This part doesn't compile in Swift 3, since your `add` method has an explicit argument.

kelvinlauKL:
Hey @GABRIELFILMSTUDIOS, are you still up for this PR?

kelvinlauKL:
Closing due to inactivity.

PR #392, opened 2017-02-26
add file ShellSortExample.swift

Moon1102:
update code.

### DIFF
--- a/Shell Sort/ShellSortExample.swift
+++ b/Shell Sort/ShellSortExample.swift
@@ -14,18 +14,18 @@ public func shellSort(_ list : inout [Int])
     
     while sublistCount > 0
     {
-        for index in 0..<arr.count{
+        for index in 0..<list.count{
             
-            guard index + sublistCount < arr.count else { break }
+            guard index + sublistCount < list.count else { break }
             
-            if arr[index] > arr[index + sublistCount]{
-                swap(&arr[index], &arr[index + sublistCount])
+            if list[index] > list[index + sublistCount]{
+                swap(&list[index], &list[index + sublistCount])
             }
             
             guard sublistCount == 1 && index > 0 else { continue }
             
-            if arr[index - 1] > arr[index]{
-                swap(&arr[index - 1], &arr[index])
+            if list[index - 1] > list[index]{
+                swap(&list[index - 1], &list[index])
             }
         }
         sublistCount = sublistCount / 2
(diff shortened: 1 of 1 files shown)


kelvinlauKL:
I'm not convinced this is an improvement over the current implementation. `shellsort.swift` looks succinct already, and follows the current explanation of it being an improvement over insertion sort. 

What sort of improvement do you want to show with your example (compared to the current implementation)?

Moon1102:
I agree with you there.
But I think, in the same way, my algorithm takes less time. Maybe my advantage is here.

part1
part2

Shell Sort/ShellSortExample.swift
```diff
@@ -0,0 +1,33 @@
...
+//  ShellSortExample.swift
+//  
+//
+//  Created by Cheer on 2017/2/26.
+//
+//
+
+import Foundation
+
+public func shellSort(_ list : inout [Int])
```
kelvinlauKL: Make sure the `:` hugs the parameter name.

Shell Sort/ShellSortExample.swift
```diff
@@ -0,0 +1,33 @@
...
+{
+    var sublistCount = list.count / 2
+    
+    while sublistCount > 0
```
kelvinlauKL: For consistency, let's have the braces trail the `while` statement.

Shell Sort/ShellSortExample.swift
```diff
@@ -0,0 +1,33 @@
...
+    {
+        for index in 0..<list.count{
```
kelvinlauKL: Add a space between `list.count` and the `{`.

Moon1102: Sure. I'll change my code to suit style guide. Thanks for your direction.

mitghi:
@Moon1102 Your implementation is much better, as it eliminated the function call and iterator overheads.

kelvinlauKL:
Looks good. Thanks @Moon1102 @mitghi !

Moon1102:
Thanks @kelvinlauKL  @mitghi

daisysomus:
@Moon1102 Your implementation has bugs. If you try this array: [64, 20, 50, 21, 72, 10, 23, -1, 4], you won't get the right answer.

Moon1102:
@daisysomus  Thank you for pointing out the mistake.

Moon1102:
image

PR #394, opened 2017-03-02
update code

Moon1102:
This radix sort faster than the original code. ^ ^
image
image


kelvinlauKL:
Looks good @Moon1102.

For future reference, could you separate these pull requests so 1 pull request accounts for 1 algorithm? 

Thanks!

Moon1102:
Okay..  
I will pay attention to it.

PR #395, opened 2017-03-07
Modified typo in README.md of QuickSort

chizcake:
I think there's typos in README.md of QuickSort :(


chizcake:
And it seems that I find another typo in line 47.
I think `equal` should be changed to `greater` and it will be 
```
This is a good split because `less` and `greater` roughly contain the same number of elements. So we've picked a good pivot that chopped the array right down the middle.
```

But I didn't commit it. Please check this line :)

kelvinlauKL:
Yup, you're right. Thanks @chizcake!

PR #397, opened 2017-03-09
Added ExpressibleByArrayLiteral to the Linked List and Tests

dsantosp12:
The Linked List data structure now conforms with the ExpressibleByArrayLiteral. For more details [here](https://developer.apple.com/reference/swift/expressiblebyarrayliteral)

### DIFF
--- a/Linked List/LinkedList.playground/Contents.swift
+++ b/Linked List/LinkedList.playground/Contents.swift
@@ -10,7 +10,7 @@ public class LinkedListNode<T> {
   }
 }
 
-public class LinkedList<T> {
+public final class LinkedList<T> {
   public typealias Node = LinkedListNode<T>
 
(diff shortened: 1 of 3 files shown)


dsantosp12:
@kelvinlauKL this is the implementation for the linked only. Before I continue applying this feature to others DS, I think a review will be good for me to have an idea of the repo's standards.

kelvinlauKL:
Yep that looks fine. You've got `ExpressibleByArrayLiteral` in it's own extension, and I don't have a problem with your treatment with `@discardableResult`. 

It's not clear whether anyone would want to subclass the linked list, but by marking the list as `final` you've guaranteed that all the methods in the Linked List to use static dispatch which should give some minor optimizations 👍 

Looks good.

PR #398, opened 2017-03-09
Dynamic Programming Knapsack Implementation

aatalyk:


aatalyk:
Dynamic Programming is one of the efficient algorithms, though it does not have specific solution for every problem there are some famous problems such as Knapsack, Coin Change.

dsantosp12:
I am new to this repo and definitely this is no extremely important, but you might want to take a look at the naming convention https://github.com/raywenderlich/swift-style-guide#naming.

kelvinlauKL:
As @dsantosp12 pointed out, we prefer the lowerCamelCase style for naming our variables.

aatalyk:
Okay, thanks I'll take it into account

aatalyk:
Do I need to add readme text right now?

kelvinlauKL:
All our contributions consists of the code and a `README` file that explains what the algorithm or data structure is about.

narrativium:
I've just reviewed the code. Before I get into it: thanks for contributing this algorithm. I think it's a useful addition, and I'm looking forward to your `README` - so I might be reviewing your code prematurely, because you might have already thought about some of the points I'm about to make.

I'm sure a lot of 'how the algorithm works' will get covered in your `README`, but from the code itself, I find it difficult to read its intent. It might be very efficient code, but it's also important here that the algorithm be easy to understand. I'm also hoping that if the code is more easily understood, it'll be easier to write about how it works.

First, you've declared several global variables which could be clearer about their role. `items`, `numberOfItems` and `values` all describe a collection of items which could be put into the knapsack. Your implementation requires that `items` and `values` are the same length, requires both arrays to have a leading `0` which doesn't get used, and `numberOfItems` should really be calculable from the size of the collection. I would suggest creating a struct (let's call it `KnapsackItem`) with a value property and a weight property. This lets you represent the items as a `[KnapsackItem]`.

Second, the global variables are declared at the top of the file but only given values at the bottom when the function gets called. I think they might work better as arguments of the knapsack function (so that it takes in information about the items, and a bag capacity, and returns an `Int`). `dp` isn't used outside the knapsack function, so should be declared inside it.

Third, the variables inside the function need to be more clearly named. Clarity is more important than brevity; a shorter variable name is easier to type but its meaning can be harder to grasp.
- `dp`. It isn't obvious to me what 'dp' stands for.
- `i` and `j` aren't just incremental counters, they're used to calculate array indices, so they need meaningful names. Perhaps replace them with `itemIndex` and `potentialCapacity`?
- I think it would be helpful if you broke down the larger calculations (I'm mostly looking at what's currently line 20) into smaller calculations.
For example, the calculation of `j - items[i]` represents the potential capacity of the bag minus the weight of item `i`, or the remaining capacity in the bag after it gets added. Take that part out of its calculation, and declare it as a constant called `remainingCapacity`. This will make it easier to tell what's being stored in the `dp` array and what it's being used for.
Better still, calculate `remainingCapacity` before the if-statement, and you can rewrite the if-statement as `if remainingCapacity < 0`. This shows the intent more clearly than `if j < items[i]`.

Last one for now, and just a small observation: the first for-loop counts from `0` to `numberOfItems`, the second for-loop counts from `0` to `capacityOfBag`. But if either value is `0`, the loop immediately proceeds to `continue`. It would be clearer if both loops started counting from `1`, and there was no test for `0`.

I hope that all makes sense. Do speak up if something's unclear to you, or if I've misunderstood something. I'm looking forward to seeing what you do with this. Thanks again for your work!

aatalyk:
Thanks, narrativium! I will consider all your suggestions. This is my first contribution so far, so basically I am new in this field. I think your suggestions are very useful and will really make the algo clear and understandable.

PR #400, opened 2017-03-13
Egg Drop Dynamic Problem

aarkalyk:
Hello!

I noticed that there aren't many dynamic programming algorithms in the repository, so I decided to contribute to the project by adding one of the most famous problems in dynamic programming "Egg Drop Problem". I'd appreciate if you could look through my code and give some feedbacks.

Thanks in advance for your attention!

### DIFF
--- a/Egg Drop Problem/EggDrop.playground/Contents.swift
+++ b/Egg Drop Problem/EggDrop.playground/Contents.swift
@@ -2,4 +2,8 @@
 
 import Cocoa
 
-eggDrop(numberOfEggs: 2, numberOfFloors: 4)
+eggDrop(numberOfEggs: 2, numberOfFloors: 2) //expected answer: 2
+eggDrop(numberOfEggs: 2, numberOfFloors: 4) //expected answer: 3
+eggDrop(numberOfEggs: 2, numberOfFloors: 6) //expected answer: 3
+eggDrop(numberOfEggs: 5, numberOfFloors: 5) //expected answer: 2
+eggDrop(numberOfEggs: 5, numberOfFloors: 20) //expected answer: 4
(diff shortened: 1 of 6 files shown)


narrativium:
More dynamic programming algorithms would be nice. If you take a look at the repository's pull request list, you can see a contribution for another dynamic programming algorithm which looks at the Knapsack problem.

I have a few comments at this point:
1) To be accepted into the repository, all contributions should have a `README` file. As a reader, I'd want to know what problem this algorithm solves, and in general terms how it solves it.
2) At the moment, sorry to say, this code wouldn't be accepted, because it doesn't compile. I don't know the egg-drop algorithm very well (see point 1) so I don't know how I'd correct it.
3) The code itself needs to make its intention clearer. This would make it easier to read, easier to understand, perhaps easier to correct, and might help you write the `README` as well.
For example: why is `eggFloor[2][1]` given a value of 1? What do the indices `i`, `j` and `k` represent? What's the significance of setting `eggFloor[i][j]` to `1000000`? In particular, consider the edge-cases - would this algorithm still work correctly if the `eggDrop` function were called and there was only 1 floor, or if there were 1000001 floors?
Use meaningful names; name your magic numbers; break complicated functions (e.g. the current line 17) into smaller, easier-to-digest steps; consider using optionals to represent situations where you don't yet know the value.
4) Provide some examples of the `eggDrop` function - ideally at the end of a playground file and in unit tests. For example, if I wrote `eggDrop(5, numberOfFloors: 3)` in my code, what value would I expect it to return to show that it was working correctly?

I hope that's fair criticism and doesn't discourage you. I look forward to seeing your revised contribution.

kelvinlauKL:
@aarkalyk Still up for this?

aarkalyk:
@narrativium Thanks for the feedback! I'll take into account all your suggestions and submit my contribution in a few days. 
@kelvinlauKL Yup!

aarkalyk:
@narrativium @kelvinlauKL Please, review my pull request. 
Thanks!

kelvinlauKL:
@aarkalyk Sorry for the wait. Going to merge this and add a few bits of polish 👍. This PR has been collecting dust long enough!

PR #401, opened 2017-03-13
A star

HellGhost:
A* is an informed search algorithm on swift


vincentngo:
@HellGhost thanks for the contribution, one of our maintainers will get to it soon!

narrativium:
Apologies for the delay @HellGhost. Thanks for contributing the implementation of A*. I'm about to review it.

I like the visual representation of the end result route - the algorithm runs pretty nicely. I like that it's taken into account that a route might not be possible, and there's a bit of demonstrable variation in the example where you make diagonal exploration optional.

My main concern with the implementation is that it's very specific. It works only for graphs which are orthogonal 2D grids measuring _exactly_ 7 nodes by 5 nodes. The two heuristics for measuring known distance to the start point and estimated distance to the end point are baked in. This makes it harder for someone else to read your code and work out what they need to change so they can apply the A* algorithm to their own problem.
Put a different way - what would have to change if the programmer was using a 5-node by 7-node grid? What if the grid was hexagonal rather than orthogonal, or 3D rather than 2D? What if the edges were all different (because the nodes represented train stations and the edges represented connections)?

The README file's quite short; it only provides a general idea of what A* is doing, and doesn't go into how you've implemented it.

To make this contribution a bit more helpful, I'd suggest you look at some of the graph-based data structures already in the SAC repository. Consider implementing an A* heuristic function in an extension to one of those existing data structures, and allowing the programmer to provide their own distance functions. So, perhaps something like this:

```
extension Graph {
  func aStarHeuristic(startNode: Node, endNode: Node, distanceBetweenNodes: (Node, Node) -> Int, estimatedDistanceToGoalNode: (Node, Node) -> Int) -> [Edge]
}
```

Hopefully it would then be easier to adapt the algorithm to other types of problem.

I hope that's not too critical. The algorithm itself is fine, even if it's not really clear why you've made certain choices (e.g. why do `AStar` and `ASNode` have to conform to `NSObject`? Would the intent be clearer if nodes were `Equatable` or `Hashable`?). I just think it needs to be a bit more general.

HellGhost:
@narrativium Hi you can use this algorithm not only with  5 by 7 this only for example and visualization, you can try to create nodes without testData, in this sample i parse test data and create objects on field which have 2 type busy or free then in AStar i select nearest nodes to start point and moves to end without any size dependencies  

And I will change inherited class AStar and ASNode, Also plane add 3D support and maybe implement additional to Graph 

I will notify you when I do this

aleph7:
This implementation of A* is O(bn^2) instead of the optimal O(bn). I'd be happy to contribute an optimal implementation. See my blog post on this: https://a-coding.com/a-data-structures/ Started implementing the data structure needed on my fork: https://github.com/aleph7/swift-algorithm-club/tree/hashed-heap

kelvinlauKL:
Hey guys, since this PR hasn't received any activity in a while, we'll continue with #633 for the A* algorithm. 

Feel free to contribute on top of #633 as we work through it. Cheers!

PR #402, opened 2017-03-16
availability check, native array loop for signal

TheIronBorn:
Couldn't run without adding an availability check


vincentngo:
@TheIronBorn looks good thanks!

PR #405, opened 2017-03-18
Minimum Spanning Tree (Kruskal's and Prim's)

Xiangxin:


vincentngo:
@Xiangxin thanks for your contributions, will have a look at this one soon!

vincentngo:
@Xiangxin Review in progress for this one!

Minimum Spanning Tree/MinimumSpanningTree.playground/Sources/Graph.swift
```diff
@@ -0,0 +1,55 @@
...
+    edgeList.append(Edge(vertex1: v1, vertex2: v2, weight: w))
+    vertices.insert(v1)
+    vertices.insert(v2)
+    if adjList[v1] == nil {
+      adjList[v1] = [(vertex: v2, weight: w)]
+    } else {
+      adjList[v1]!.append((vertex: v2, weight: w))
+    }
+
+    if adjList[v2] == nil {
```
Xiangxin: Thanks for pointing it out! I felt a bit uncomfortable about this when I coded this as well.
There's a problem with if let
screen shot 2017-04-25 at 3 40 59 pm
I was thinking of something like 
```
adjList[v1] = adjList[v1] ?? []()
adjList[v1]?.append((vertex: v2, weight: w))]
```
as well but it's not as efficient since value type assignment does copy the entire value.

I removed the force unwrapping and made it a bit more Swift-ier.

Xiangxin:
@vincentngo Thanks for reviewing!

Minimum Spanning Tree/MinimumSpanningTree.playground/Sources/Graph.swift
vincentngo: @Xiangxin I was wondering the same, but after talking to some of my peers, it will not internally make a new copy unless adjList[v1] was mutated. You can go with either  nil coalescing or what you originally had with adjList[v1] == nil check. I prefer nil checks over if let _ = adjList[v1] since the intent is more clear. Thanks!

vincentngo:
@Xiangxin thanks for the awesome contribution!

PR #408, opened 2017-03-21
Updated red-black tree

uteschiehlen:
In the current implementation of the red-black tree the deletion algorithm is not working correctly. 
For example if you insert 1 and then 2. And then delete 1, the root is not correctly updated.
Example:
`
testTree.insert(1)
testTree.insert(2)
`
generates the tree: (null<-|1|->(null<-(2)->null))
Then:
`
testTree.delete(1)
`
has the result: (null<-|1|->(null<-|2|->null))
which is not correct, it should be (null<-|2|->null)

I've re-written the insert and delete operation according to the pseudo algorithm provided in "Introduction to Algorithms" by CLRS.


vincentngo:
@uteschiehlen there seems to be some merge conflicts, could you please review? Thanks.

uteschiehlen:
@vincentngo I've resolved the merge conflicts. As I've rewritten the whole RedBlackTree class, I deleted the old implementation including the playground. I hope this is ok, as otherwise we would have had two implementations of the tree.

vincentngo:
@uteschiehlen sounds good, thank you

vincentngo:
@uteschiehlen just giving you a heads up will finish the review by this week. Sorry!

Red-Black Tree/RedBlackTree.playground/Sources/RedBlackTree.swift
```diff
@@ -0,0 +1,751 @@
...
+}
+
+// MARK: - RBNode
+
+public class RBTreeNode<Key: Comparable>: Equatable {
+  public typealias RBNode = RBTreeNode<Key>
+  
+  fileprivate var color: RBTreeColor = .black
+  fileprivate var key: Key?
+  fileprivate var isNullLeaf = true
```
vincentngo: Is it necessary to have a variable for this? isNullLeaf just checks to see if it's key, leftchild,rightChild, and parent is nil right? If we could remove this, that's one less variable to maintain. We could create a computed property, or func tion to check if it's a nullLeaf maybe?

Red-Black Tree/RedBlackTree.playground/Sources/RedBlackTree.swift
```diff
@@ -0,0 +1,751 @@
...
+}
+
+// MARK: - Finding maximum and minimum value
+
+extension RedBlackTree {
+  /*
+   * Returns the minimum key value of the whole tree
+   */
+  public func minValue() -> Key? {
+    if let minNode = root.minimum() {
```
vincentngo: this is fine, something a little cleaner would be to use a guard let statement. 

E.g. 
```
guard let minNode = root.minimum()  else { return nil }
return minNode.key
```

Red-Black Tree/RedBlackTree.playground/Sources/RedBlackTree.swift
```diff
@@ -0,0 +1,751 @@
...
+       return minNode.key
+    } else {
+      return nil
+    }
+  }
+  
+  /*
+   * Returns the maximum key value of the whole tree
+   */
+  public func maxValue() -> Key? {
```
vincentngo: Could also use the guard let pattern instead.

vincentngo:
@uteschiehlen what's the status on this based on review comments? Thanks.

uteschiehlen:
@vincentngo I've incorporated your reviews. You are correct, we do not need isNullLeaf to be a variable, so I changed it to a computed property, thanks for pointing this out.

PR #413, opened 2017-03-24
Fix links to related articles

jkrmr:
Fixes links to Linked List, Ring Buffer, and Priority Queue directories.

[Rendered](https://github.com/jkrmr/swift-algorithm-club/blob/a54e722b035ffc5eeba63fcd5b1648e56f732ec0/Queue/README.markdown):
_____________________________________________
## See also

There are many other ways to create a queue. Alternative implementations use a [linked list](../Linked%20List/), a [circular buffer](../Ring%20Buffer/), or a [heap](../Heap/). 

Variations on this theme are [deque](../Deque/), a double-ended queue where you can enqueue and dequeue at both ends, and [priority queue](../Priority%20Queue/), a sorted queue where the "most important" item is always at the front.

### DIFF
--- a/Queue/README.markdown
+++ b/Queue/README.markdown
@@ -255,8 +255,8 @@ The `nil` objects at the front have been removed and the array is no longer wast
 
 ## See also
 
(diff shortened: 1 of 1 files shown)


kelvinlauKL:
Looks good! Thanks

PR #414, opened 2017-03-25
URL encode links in Markdown files

jkrmr:
There are a ton of broken markdown links in the READMEs throughout this repo because the files and directories they refer to have spaces in their names but these aren't HTML encoded in their links. This commit just replaces spaces with %20 to fix these.

before:
1b-before

after:
1a-after

before:
2a-before

after:
2b-after


kelvinlauKL:
Cheers!

PR #415, opened 2017-03-26
Fix binary search tree link

huntermonk:
Just need to convert spaces into %20 for relative URLs
with spaces to work.


kelvinlauKL:
Cheers!

PR #416, opened 2017-03-27
update ShellSortExample.swift

Moon1102:
fix bug


Shell Sort/ShellSortExample.swift
```diff
@@ -8,24 +8,25 @@
...
-   
-    while sublistCount > 0 {
-        
-        for index in 0..<list.count {
-           
-            guard index + sublistCount < list.count else { break }
+    
+    while sublistCount > 0
+    {
+        for var index in 0..<arr.count{
```
kelvinlauKL: Where is `arr` declared?

Moon1102: I updated the code and checked the format. Hope has fixed the mistake of carelessness.

kelvinlauKL:
Looks good, thanks @Moon1102!

PR #426, opened 2017-04-03
Minimum Coin Change Problem - Dynamic Programming in Swift

JacopoMangiavacchi:
Minimum Coin Change problem algorithm implemented in Swift comparing dynamic programming algorithm design to traditional greedy approach.


kelvinlauKL:
Hey!

Just noticed that you're submitting this under the Apache license. We're currently operating under the MIT license. Could you please switch the licenses?

JacopoMangiavacchi:
Hi Kelvin, just updated to the MIT license.  Cheers, Jacopo

narrativium:
Hello there. I'm just reviewing the code for this algorithm - thanks for contributing it. I especially like that there's a greedy algorithm present for comparison, with a simple example of how it fails.

I'm happy with the algorithms in general - the code's quite readable - but I have a few comments about the dynamic programming algorithm's implementation. Some of these are more about style, some more about correctness, and some might involve edge cases, but at the moment I don't know which of these should require a change in the code, and which might be best served with a comment or note in the README.

Coin Denomination Inputs:
1) The data structure assumes all `Int`s are acceptable coin denominations, but only positive integer values should be. The algorithms might not finish if zero or negative integer values were accepted. It's an edge case, but asserting that all denominations be positive or filtering out all the non-positive denominations might be sensible.
2) The algorithms assume that any sum can be made with any set of denominations, but this is a false assumption. For example, if I wanted a combination of coins which made 21 cents, and only had 20-cent coins available, both algorithms would return a solution of a list containing a single '20'. If I made a 7-cent coin available, the dynamic programming algorithm would find the solution of three 7-cent coins (which the greedy algorithm wouldn't), but would reject it in favour of the single 20-cent coin 'solution' because it has fewer coins. Should the algorithms be rewritten to return a `[Int]?`, only returning a solution which exactly add up to the requested total, or nil if it can't be done?

The `cached` property:
Since this property is part of the type rather than the algorithm, the struct will 'learn' from multiple calls of the dynamic programming algorithm. This isn't a bad thing and you do explicitly mention it, but if you wanted the dynamic programming algorithm to start a fresh cache each time, you can re-write the algorithm as a non-recursive function which contains a recursive function (remember that in Swift, functions can be nested inside other functions).

The `changeDynamic` function:
I think this could benefit from a little polish. For example: on what's currently line 54, you can declare `var potentialChange = [coin]`. On line 58, you're testing if `value` is positive, even though you already tested for this on line 40 and haven't changed its value since.

The test file:
This doesn't actually test anything, it simply runs both the algorithms a hundred times based on one set of denominations. A demonstration like this probably works better in a playground.
Unit test files would be useful, though. Testing both algorithms on your [1, 3, 4] -cent denominations, 6-cent example would be good to demonstrate.

I hope these comments make sense. As I said, I'm not sure all the comments require changes to what you've done, but if you think based on these that your contribution can be improved then I look forward to seeing it.

JacopoMangiavacchi:
Hi narrativium, thank you very much for your review.  I agree on all of your feedbacks and pushed back a fix for everything.  Cheers, Jacopo

JacopoMangiavacchi:
Hi @narrativium , did you had a chance to review the changes I've applied based on your suggestions above ?

narrativium:
@JacopoMangiavacchi: I have had a chance, and it looks good, but I've just looked over it one more time and one thing still needs a tweak as it's not there yet.

The dynamic algorithm (in `MinimumCoinChange.swift`, lines 71-78, the `_changeDynamic` inner function) is still sorting potential change arrays by size before it checks whether the array adds up to the desired value.
So it's prioritising smaller numbers of coins which don't hit the target, over larger numbers of coins which do.

Can you modify it so that it filters the potential change arrays for only the ones which add up to the exact value, and then pick the smallest from those? And if there is such an array, cache it as you are doing.

The rest is fine, I think we're pretty close to merging this contribution. Thanks!

JacopoMangiavacchi:
Thanks @narrativium Another very good one.  Done !

JacopoMangiavacchi:
Just last minute cleanup: remove no more needed temporary variable for the winner array

narrativium:
@JacopoMangiavacchi: sorry for the delay in reviewing this.

I've just had another look at it. It looks good, and I'm merging it in.

Issue #317, opened 2016-12-17
Data Structures / Algorithms Wishlist


kelvinlauKL:
### Updates Wishlist
- [ ] The Galil and Good Suffix Rules for Boyer-Moore - @Frizlab - #312

kelvinlauKL:
Closing to follow hacktoberfest's conventions.

PR #429, opened 2017-04-04
Fixing a bug in the Linked List logic

m-alani:
This PR address 2 issues in the Linked List files:

- The playground had few wrong comments about the expected output of the statements
- The implementation of the Linked List has a logical error where a `Node` is used as is when passed to the `append` or `insert` functions. The correct way of implementing such functions is to ALWAYS copy the node's value into a new node, and then insert the new node into the data structure.

Without the second fix, a simple statement like `list.append(list.first)` would crash the playground if the list has more than 1 node (`last.next` won't be `nil` in this case)

### DIFF
--- a/Linked List/LinkedList.playground/Contents.swift
+++ b/Linked List/LinkedList.playground/Contents.swift
@@ -73,7 +73,8 @@ public final class LinkedList<T> {
     self.append(newNode)
   }
   
-  public func append(_ newNode: Node) {
+  public func append(_ node: Node) {
+    let newNode = LinkedListNode(value: node.value)
     if let lastNode = last {
       newNode.previous = lastNode
       lastNode.next = newNode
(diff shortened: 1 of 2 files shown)


kelvinlauKL:
Good catch on the comments. I have a few reservations regarding the changes you want to make to the `append` and `insert` logic. 

My conjecture is that these methods were written to allow other linked lists to be appended directly. For example:

```
list1: A -> B -> C

list2: D -> E -> F

list1.append(list2.first!): A -> B -> C -> D -> E -> F 
```

Using your implementation, `list1.append(list2.first!)` will yield `A -> B -> C -> D` since you only copied the value of the first. If that's the intention, you could use the other `append` function by calling `list1.append(list2.first!.value)`.

That being said, I think you're on to something. I think there's room to improve on, especially on expressiveness and clarity of the interface. 

Let me know what you think!

m-alani:
I see your point @kelvinlauKL.

Judging by other implementations of linked-list, I think it'd make more sense to overload the `insert` and `append` methods to handle a linked-list as an argument, so my suggestion is to keep the `insert` and `append` as is in the PR when taking a `node` as an argument, and if you (and the community) agree, I can add 2 more implementations (with few test cases) to have `insert` and `append` accept a `LinkedList` type and insert that whole list at the specified position (or to the end of the list).

Looking forward to hearing some feedback.

kelvinlauKL:
That sounds good. I've added that to #317, for visibility. Thanks!

PR #432, opened 2017-04-05
Revisions in README.markdown

Parva9eh:
I have revised your document to be simplified and concise.


Parva9eh:
@vincentngo Thanks.

PR #441, opened 2017-04-15
Dynamic Programming. 0-1 Knapsack Problem.

aatalyk:
Hi, there. I finished my implementation of Knapsack Problem in Swift. I close old request because I accidentally deleted all my files and didnt save them. I rewrote the algo. Now It would be good if review my solution and give some feedback, or if it is okay can you accept my request. Thanks!


Dynamic Programming/0-1 Knapsack/Knapsack.playground/Contents.swift
```diff
@@ -0,0 +1,51 @@
...
+    init(weight: Int, value: Int) {
+        self.weight = weight
+        self.value = value
+    }
+}
+
+var capacityOfBag = Int()
+var knapsackItems = [KnapsackItem]()
+
+func knapsack() -> Int {
```
kelvinlauKL: This was somewhat confusing to read. Instead of having this free floating function for preparing the knapsack, let's create a class that models the knapsack. I've came up with this:

```
public final class Knapsack {
  public let capacity: Int

  // capacityOfBag would be a property of this class, passed in via the initializer
  public init(weightCapacity: Int, itemTypes: [KnapsackItem]) {
    capacity = weightCapacity

    // do something with the KnapsackItems
  }
}
```

I can see a few ways to model the solution. We could configure this Knapsack with the optimal `KnapsackItem` combination right in the initializer, or we can create a function that takes a set of `KnapsackItem` values and returns the optimal combination.

Let me know if that makes sense.

Dynamic Programming/README.markdown
```diff
@@ -0,0 +1,190 @@
...
+## Problem
+
+Given `KnapsackItem` items with corresponding weigths and values find the maximum total value in the knapsack by putting `KnapsackItem` items in a knapsack of capacity `capacityOfBag`. 
+
+In other words, given array `knapsackItems` which represent values and weights associated with `KnapsackItem` items respectively. Also given an integer `capacityOfBag` which represents knapsack capacity, find out the maximum value subset of `knapsackItems` such that sum of the weights of this subset is smaller than or equal to `capacityOfBag`. You cannot break an item, either pick the complete item, or don’t pick it ([0-1 Knapsack Problem](http://www.geeksforgeeks.org/dynamic-programming-set-10-0-1-knapsack-problem/)).
+
+`KnapsackItem` is `NSObject` class with two properties such as the weight and the value.
+
+```swift
+class KnapsackItem: NSObject {
```
kelvinlauKL: I mentioned this in another comment, but can we do a struct instead?

```
struct KnapsackItem {
  let weight: Int
  let value: Int
}
```

aatalyk: Thank you, kelvinlauKL. I will take into account your suggestions.

kelvinlauKL:
Closing due to inactivity. Feel free to open it up again if you still wish to add this in.

PR #446, opened 2017-04-26
New algorithm ArrayLimitedQueue added

SeRG1k17:
### Checklist

### Description
This collection is convenient because it combines the functionality of other data structures. It also has a user-friendly interface and dynamic properties. The best example to use is the table(UITableView) with the "opening" sections.

### DIFF
--- a/ArrayLimitedQueue/ArrayLimitedQueue.swift
+++ b/ArrayLimitedQueue/ArrayLimitedQueue.swift
@@ -1,0 +1,182 @@
+//
+//  ArrayLimitedQueue.swift
+//
+//  Created by Sergey Pugach on 15.04.17.
+//  Copyright © 2017 Sergey Pugach. All rights reserved.
+//
+
+/*
+Array Limited Queue is a collection that has the features
+ of a regular array (only get + limited size), queues and sets.
+ The way you value and compare items can be user-defined.
+*/
+
+import Foundation
+
+public struct ArrayLimitedQueue<T: Comparable> : CustomStringConvertible {
+    
+    //The maximum number of items in the collection. Can be changed after
+    public var maxSize: Int = 1 {
+        didSet {
+            let sizeDiff = internalArray.count - maxSize
+            if 0 < sizeDiff && sizeDiff <= internalArray.count {
+                internalArray.removeFirst(sizeDiff)
+            }
+        }
+    }
+    /*
+    This property determines which elements can be in the collection.
+    After the change from true to false. Negative values will be deleted
+    Note: When using custom types, you need to set the zeroValue.
+     */
+    public var zeroValue:T?
+    public var positiveValues: Bool = false {
+        didSet {
+            if positiveValues {
+                
+                if let zero = zeroValue {
+                    internalArray = internalArray.filter{$0 > zero}.map{ $0 }
+                    
+                } else if let defaultZero = 0 as? T {
+                    internalArray = internalArray.filter{$0 > defaultZero}.map{ $0 }
+                    
+                } else {
+                    fatalError("A zeroValue is not setted")
+                }
+            }
+        }
+    }
+    
+    /*
+     Specifies whether to remove duplicates from the collection. 
+     When a copy is found, it leaves the last value.
+     */
+    public var deleteExisting = true {
+        didSet {
+            if deleteExisting {
+                internalArray = internalArray.reversed().reduce([]){$0.contains($1) ? $0 : $0 + [$1]}.reversed()
+            }
+        }
+    }
+    
+    func removeDuplicates(accumulator: [T], element: T) -> [T] {
+        return accumulator.contains(element) ? accumulator : accumulator + [element]
+    }
+    
+    public var array: [T] {
+        return internalArray
+    }
+    private var internalArray = [T]()
+    
+    public var isEmpty: Bool {
+        return count == 0
+    }
+    
+    // Returns the number of elements in the collection.
+    public var count: Int {
+        return internalArray.count
+    }
+    
+    // Returns the 'maximum' or 'largest' value in the collection.
+    public var maxValue: T? {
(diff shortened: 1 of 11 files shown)


ArrayLimitedQueue/README.md
```diff
@@ -0,0 +1,276 @@
+# ArrayLimitedQueue
+
+Array Limited Queue is a collection that has the features of a regular array (only get + limited size), queues and sets. The way you value and compare items can be user-defined.
+
+The Ordered Set data type is a hybrid of:
+- a [Queue](https://en.wikipedia.org/wiki/Queue_(abstract_data_type)), is a list where you can only insert new items at the back and remove items from the front. This ensures that the first item you enqueue is also the first item you dequeue. First come, first serve!
+- a [Set](https://en.wikipedia.org/wiki/Set_%28mathematics%29), a collection of unique items where the order does not matter, and
+- a [Sequence](https://en.wikipedia.org/wiki/Sequence), an ordered list of items where each item may appear more than once.
```
SeRG1k17: @kelvinlauKL In the near future I will complete this work(time of diploma :)  ). Do you really think that this is a useful algorithm?

kelvinlauKL: A problem that could use `queue` and `set` semantics could make use of this. The thing with data structures is that everything has a use, even if some use cases are more niche than others.  

I took another look at the code and from a performance standpoint, there could be improvements that you can make.

For instance, `maxSize` trims the array by using `removeFirst`. Each time `removeFirst` is called, your array needs to shift all the remaining elements on the left - an **O(n)** operation. (Though if you changed the internal storage type from an array to something else, it won't be called a **Array**LimitedQueue anymore...). A Linked List could be more efficient.

In the `deleteExisting` observer, this is pretty heavy work:

```
internalArray = internalArray.reversed().reduce([]){$0.contains($1) ? $0 : $0 + [$1]}.reversed()
```

- `reversed` is **O(n)**
- `reduce` is **O(n)**
- `contains` is **O(n)**
- `reversed` is **O(n)**

Like I mentioned above, I like the functionality this data structure can provide. The interface seems functionally appealing, but very unswift like at the moment. 

Writing this:

```
maxSize = 10
```

not only defines how many elements this data structure can hold, but also removes anything over that limit from the front of the queue. That makes sense, but feels a bit weird at the same time.

SeRG1k17:
@kelvinlauKL Hi, I'm here again and intend to finish this job. I fixed the difficult one in the work in the DexDesign business. I do not think it's worth changing the type of internal storage, because in this algorithm there is access to the center and the array is most convenient for this. Let's continue our discussion to complete the work.

kelvinlauKL:
Hey @SeRG1k17, great to hear from you again.

I took another look at the code, and the amount of `removeFirst` calls strongly suggests to me that an internal `LinkedList` will be a huge win in terms of performance. 

That being said, let's revisit the internal storage topic a bit later. For now, I want to focus on semantics. To start, could you update the code across your files? I'm looking at the files for this PR and noticed that there are two different implementations of `ArrayQueueLimited`. 

Looking forward to getting this merged!

SeRG1k17:
@kelvinlauKL check it now

ArrayLimitedQueue/ArrayLimitedQueue.swift
```diff
@@ -0,0 +1,173 @@
...
+                            print("remove :\(internalArray[j])")
+                            internalArray.remove(at: j)
+                        }
+                    }
+                }
+            }
+        }
+    }
+    
+    func removeDuplicates(accumulator: [T], element: T) -> [T] {
```
kelvinlauKL: This seems unused. What's the purpose of this method?

ArrayLimitedQueue/ArrayLimitedQueue.swift
```diff
@@ -0,0 +1,173 @@
...
+        return accumulator.contains(element) ? accumulator : accumulator + [element]
+    }
+    
+    public var array: [T] {
```
kelvinlauKL: We could probably combine `array` and `internalArray` to a single property, since it seems like the only purpose of having the two is to prevent others from tampering with the internal storage:

```
public private(set) array: [T] = []
```

This will prevent users from manipulating the array directly, but still allow them to read from it.

ArrayLimitedQueue/ArrayLimitedQueue.swift
```diff
@@ -0,0 +1,173 @@
...
+                internalArray.removeFirst(sizeDiff)
+            }
+        }
+    }
+    /*
+     This property determines which elements can be in the collection.
+     After the change from true to false. Negative values will be deleted
+     Note: When using custom types, you need to set the zeroValue.
+     */
+    public var zeroValue:T? = 0 as? T
```
SeRG1k17: removed

ArrayLimitedQueue/ArrayLimitedQueue.swift
SeRG1k17: removed

SeRG1k17:
@kelvinlauKL I reworked the algorithm into a linked list. BUT I faced one problem, maybe you will help me solve it:
i need use it func for two functions in my algo
```
    //Returns the index of an item if it exists
    public func indexOf(item: T) -> Int? {
        return list.index(of: item)
    }
```
But linkedList algorithm cant compare two nodes

ArrayLimitedQueue/ArrayLimitedQueue.swift
SeRG1k17: I do not know how it turned out there in my version of the file this is long gone

kelvinlauKL:
Hey @SeRG1k17 

Sorry for dropping the ball on this one. Could you check and let me know the error message for that line?

kelvinlauKL:
Closing due to inactivity.

PR #452, opened 2017-04-30
switch from local removeFirst copy to iteration for pairing

TheIronBorn:
### Checklist


Select Minimum Maximum/Maximum.swift
```diff
@@ -3,13 +3,12 @@
  */
 
 func maximum<T: Comparable>(_ array: [T]) -> T? {
-    var array = array
     guard !array.isEmpty else {
```
kelvinlauKL: Since we're force unwrapping `array.first` due to this `guard` statement's guarantee, let's merge the two:

```
// guards against the empty array and also extracts the first element of the array
guard let maximum = array.first else { return nil }
```

TheIronBorn:
Oh yeah that's fine.  I'll implement in a minute.

kelvinlauKL:
Looks great! Cheers

PR #499, opened 2017-07-19
Encode and Decode A Binary Tree

remlostime:
### Checklist

### Description

Encode and Decode A Binary Tree

### DIFF
--- a/Encode and Decode Tree/EncodeAndDecodeTree.swift
+++ b/Encode and Decode Tree/EncodeAndDecodeTree.swift
@@ -8,97 +8,88 @@
 
 import Foundation
 
-public class TreeNode {
-    public var val: String
-    public var left: TreeNode?
-    public var right: TreeNode?
-
-    public init(_ val: String, left: TreeNode? = nil, right: TreeNode? = nil) {
-        self.val = val
-        self.left = left
-        self.right = right
-    }
+protocol BinaryNodeEncoder {
+  func encode<T>(_ node: BinaryNode<T>?) throws -> String
 }
 
-public class EncodeAndDecodeTree {
-    private var encodeStr = ""
-    private var ret: [String] = []
-
-    public init() {}
-
-    public func encode(_ root: TreeNode?) -> String {
-        encodeIt(root)
-
-        return encodeStr
-    }
-
-    public func decode(_ data: String) -> TreeNode? {
-        var s = data
-        while (s.contains(" ")) {
-            let index = s.index(of: " ")
-            let endIndex = s.index(before: index)
-            let element = s[s.startIndex...endIndex]
-            ret.append(element)
-            let range = ClosedRange(uncheckedBounds: (lower: s.startIndex, upper: index))
-            s.removeSubrange(range)
-        }
-
-        if ret.count == 0 {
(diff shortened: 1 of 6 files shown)


Encode and Decode Tree/EncodeAndDecodeTree.swift
```diff
@@ -0,0 +1,104 @@
...
+
+    public init(_ val: String, left: TreeNode? = nil, right: TreeNode? = nil) {
+        self.val = val
+        self.left = left
+        self.right = right
+    }
+}
+
+public class EncodeAndDecodeTree {
+    private var encodeStr = ""
```
kelvinlauKL: FYI, this is a class member variable and violates one of the rules for this problem:

> Do not use class member/global/static variables to store states. Your serialize and deserialize algorithms should be stateless.

kelvinlauKL:
@remlostime I did an editing pass on the `README` file and rewritten some code to be more Swifty. 

I haven't rewritten the code to match the `README` yet, are you interested in doing that?

remlostime:
@kelvinlauKL Thanks for the awesome editing so many times! Really appreciate that! Will update the PR this week.

remlostime:
@kelvinlauKL I looked into updated doc, I feel put the `encode` and `decode` func into `Node` is a little bit weird for me. Imk, the node should have no knowledge about tree funcs. It's much tidy in the code level to put everything into `Node`, but it seems that in the model level, we should split the job from `Node`. Thoughts?

kelvinlauKL:
@remlostime Good points! This is topic is particularly interesting to talk about. I've been reading [Clean Architecture](https://www.amazon.com/Clean-Architecture-Craftsmans-Software-Structure/dp/0134494164) lately, and the author writes about the **Single Responsibility Principle** (SRP).

SRP states that **a module should be responsible to one, and only one, actor**. I *feel* that implies that encoding and decoding logic should reside in a separate module (a different file, or namespace).

However, in the context of encoding and decoding models, the Swift team/community is encouraging us to have the models know about its own encoding and decoding operations.

As of Swift 4.0, `Encodable` and `Decodable` protocols have been added to the stdlib, and models are encouraged to adopt it to make use of decoder types such as the new `JSONDecoder` objects. 

In light of this discussion, I think we should adopt `Encodable` and `Decodable` protocols for the `BinaryNode` type to be consistent with the new idioms of the Swift language.

What do you think?

kelvinlauKL:
Related info to encoding and decoding custom types: https://developer.apple.com/documentation/foundation/archives_and_serialization/encoding_and_decoding_custom_types

kelvinlauKL:
@remlostime I made a mistake, `Encodable` and `Decodable` don't actually hold the encoding and decoding functions. 

Those functions actually belong to the `Encoder` and `Decoder` protocols.

remlostime:
@kelvinlauKL Thanks for bringing this awesome resource! That looks great to me! I can update the code to match the doc.

remlostime:
@kelvinlauKL Updated the code to sync up the readme. PTAL.
Sorry for late. :)

kelvinlauKL:
Thanks!

PR #515, opened 2017-07-26
Correct typo

specialfor:
Algorithm doesn't work because of usage of incorrect set of vertices.


crabman448:
Thank you very much! Sorry, that I did not notice this bug earlier.

PR #581, opened 2017-08-05
Add hashed heap data structure

aleph7:
### Checklist

### Description

Even though there already is a PR for an A* implementation, that PR implements it in O(bN^2) instead of the optimal O(bn). This data structure is the necessary building block for an optimal implementation. It is just a variation on the existing regular heap, but adding a hash map for fast lookups by value.

### DIFF
--- a/Hashed Heap/HashedHeap.swift
+++ b/Hashed Heap/HashedHeap.swift
@@ -31,13 +31,13 @@ public struct HashedHeap<T: Hashable> {
     /// - Complexity: O(n)
     public init(array: [T], sort: @escaping (T, T) -> Bool) {
         isOrderedBefore = sort
-        buildHeap(fromArray: array)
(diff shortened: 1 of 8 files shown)


Hashed Heap/HashedHeap.swift
```diff
@@ -0,0 +1,239 @@
...
+    @inline(__always)
+    func leftChildIndex(ofIndex i: Int) -> Int {
+        return 2*i + 1
+    }
+
+    /// Returns the index of the right child of the element at index i.
+    ///
+    /// - Note: this index can be greater than the heap size, in which case there is no right child.
+    @inline(__always)
+    func rightChildIndex(ofIndex i: Int) -> Int {
```
aleph7: Great suggestions, fixed.

PR #582, opened 2017-08-06
[Swift 4] Singly Linked List with value semantics, copy-on-write and …

newboadki:
…collection protocol conformance.

### Checklist

### Description

 

This is an implementation of a Singly Linked List that aims to contribute with a more idiomatic Swift Implementation of a linked list:
- Value Semantics.
- Copy on write.
- Conformance to the Collection protocol.

By adopting the patterns that the structures built-in in the language's standard Library we can achieve consistency as well as take advantage of the efficiency strategies present in the Swift's Standard Library.


Singly Linked List/SinglyLinkedList.swift
```diff
@@ -0,0 +1,581 @@
...
+    
+    
+    
+    // MARK: INITIALIZERS
+    
+    /// Creates a list with the given node.
+    /// NOTE: This method can break value semantics by accepting a node.
+    ///
+    /// - Parameter head: First node
+    internal init(head: SinglyLinkedListNode<T>)
```
newboadki: Sure, removed.

Singly Linked List/SinglyLinkedList.swift
```diff
@@ -0,0 +1,581 @@
...
+    {
+        self.storage = IndirectStorage()
+        self.append(node: head)
+    }
+    
+    
+    /// Creates a list with a single element
+    ///
+    /// - Parameter value: element to populate the list with
+    public init(value: T)
```
newboadki: Done.

Singly Linked List/SinglyLinkedList.swift
```diff
@@ -0,0 +1,581 @@
...
+    }
+    
+    
+    
+    // MARK: LOOP DETECTION
+    
+    /// A singly linked list contains a loop if one node references back to a previous node.
+    ///
+    /// - Returns: Whether the linked list contains a loop
+    public func containsLoop() -> Bool
```
newboadki: Removed.

Singly Linked List/SinglyLinkedList.swift
```diff
@@ -0,0 +1,581 @@
...
+        self.head = self.head?.next
+        return result
+    }
+}
+
+
+
+// MARK: - SEQUENCE -
+
+extension SinglyLinkedList : Sequence
```
newboadki: Removed.

Singly Linked List/SinglyLinkedList.swift
```diff
@@ -0,0 +1,581 @@
...
+    public func index(after idx: Index) -> Index {
+        return SinglyLinkedListIndex<T>(node: idx.node?.next, tag: idx.tag+1)
+    }
+}
+
+
+
+// MARK: - QUEUE -
+
+extension SinglyLinkedList : Queue
```
newboadki: Removed.

newboadki:
Thank you for your comments. I have addressed your suggestions and the code styling issues. In regards of adding the following three points to the existing implementation of LL:

1. CoW
The reason why I have not modified the existing implementation is because changing the semantics of an existing class felt very intrusive. Doing so would mean changing LL to be a struct. This could introduce bugs to existing client code, since copies of the value would start taking place potentially unnoticed by the developer.

If the goal is to develop a framework of data structures, then It would be good to make a design decision on the semantics that structures to be develop would have and stick to it. If the goal is purely didactic and of divulgation of knowledge then I personally think having a variety of semantics and examples can be useful to the reader. It would be then a matter of organising the folders intuitively. Maybe, we could group related implementations under the same folder, i.e Linked Lists? Inside we could differentiate according to semantics, or simply have this in the name? How do you feel about that?

2. Collection conformance to the existing LL.
Can do.

3. ExpressibleByArrayLiteral to the existing LL.
Can do.

kelvinlauKL:
Good points. Sorry for the delayed response, been on vacation and got a big backlog of prs. This is still on my todo list!

PR #584, opened 2017-08-06
Ring Buffer Refinements

alessandro-martin:
### Checklist

### Description

- Added array private extension to wrap indexes;
- Added Sequence conformance;
- Simplified read and write implementations;
- When reading, value is nilled-out.

### DIFF
--- a/Ring Buffer/RingBuffer.playground/Contents.swift
+++ b/Ring Buffer/RingBuffer.playground/Contents.swift
@@ -9,24 +9,20 @@ public struct RingBuffer<T> {
     array = [T?](repeating: nil, count: count)
   }
 
+  /* Returns false if out of space. */
+  @discardableResult
   public mutating func write(_ element: T) -> Bool {
-    if !isFull {
-      array[writeIndex % array.count] = element
-      writeIndex += 1
-      return true
-    } else {
-      return false
-    }
+    guard !isFull else { return false }
+    defer { writeIndex += 1 }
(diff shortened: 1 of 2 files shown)


narrativium:
I'd like to merge this, but can I ask you to change a few lines to comply with the style guidelines?
I'm looking specifically at some of your uses of `defer`, where you've put several code statements onto a single line separated by semi-colons. We'd prefer each code statement to be on a new line.

Other than that it's fine and I'm happy to merge the pull request.

hollance:
Personally I find those functions with defer in them much less readable than the originals. It looks like an effort to use fancy Swift stuff like guard and defer where the if statement works just as well and actually does things in the order that they are stated. I'm all for improving the code but these are pointless changes.

hollance:
I do like the wrapped subscript, though. :-)

alessandro-martin:
@narrativium I have made the changes you requested.

alessandro-martin:
@hollance I can assure you I am not trying to be fancy or clever; I find, in general, that the use of guard and defer in place of if/else constructs improves both the flow and the readability.

kelvinlauKL:
I'm a big fan of `guard`, but I'm not sure about the `defer` usage in this situation. Is `defer` called if `guard` fails in the changes you've made? 

`defer` is good if you needed to do the same method calls in multiple exit points of the function. In this case, it's functionally equivalent whether or not you have the `defer` statement there.

PR #610, opened 2017-09-05
Improvements to Multiset

simonwhitaker:
### Checklist

### Description

- Improves the README with better examples
- Adds a constructor for instantiating the set with a Collection of Hashasbles

### DIFF
--- a/Multiset/Multiset.playground/Sources/Multiset.swift
+++ b/Multiset/Multiset.playground/Sources/Multiset.swift
@@ -7,18 +7,24 @@
 
 import Foundation
 
-public struct Multiset<Element: Hashable> {
-  private var storage: [Element: UInt] = [:]
+public struct Multiset<T: Hashable> {
+  private var storage: [T: UInt] = [:]
   public private(set) var count: UInt = 0
 
   public init() {}
 
-  public mutating func add (_ elem: Element) {
+  public init<C: Collection>(_ collection: C) where C.Element == T {
+    for element in collection {
+      self.add(element)
+    }
+  }
+
+  public mutating func add (_ elem: T) {
     storage[elem, default: 0] += 1
     count += 1
   }
 
-  public mutating func remove (_ elem: Element) {
+  public mutating func remove (_ elem: T) {
     if let currentCount = storage[elem] {
       if currentCount > 1 {
         storage[elem] = currentCount - 1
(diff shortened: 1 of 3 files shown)


Multiset/Multiset.playground/Sources/Multiset.swift
```diff
@@ -7,18 +7,24 @@
 
 import Foundation
 
-public struct Multiset<Element: Hashable> {
-  private var storage: [Element: UInt] = [:]
+public struct Multiset<T: Hashable> {
+  private var storage: [T: UInt] = [:]
```
simonwhitaker: Changing the type parameter name was necessary, because Xcode 9 was failing to compile the file otherwise with this change. My assumption is that it was struggling with L16, which otherwise read:

    public init<C: Collection>(_ collection: C) where C.Element == Element {

Regardless of Xcode's problems, this change also makes that line easier to read.

Multiset/Multiset.playground/Sources/Multiset.swift
```diff
@@ -7,18 +7,24 @@
...
   public private(set) var count: UInt = 0
 
   public init() {}
 
-  public mutating func add (_ elem: Element) {
+  public init<C: Collection>(_ collection: C) where C.Element == T {
```
kelvinlauKL: Great addition. I propose we also add a little Swift syntactical sugar by adopting the `ExpressibleByArrayLiteral` protocol:

```
extension Multiset: ExpressibleByArrayLiteral {
  public init(arrayLiteral elements: T...) {
    self.init(elements) // using the `Collection` initializer you created
  }
}
```

This allows us to write code like this:

```
let multiset: Multiset<String> = ["Cacti", "Tactical"]
```

Multiset/README.markdown
```diff
@@ -32,7 +50,7 @@ Here's the essence of it:
 
 ``` swift
 public struct Multiset<Element: Hashable> {
-  public private(set) var storage: [Element: UInt] = [:]
+  private var storage: [Element: UInt] = [:]
```
kelvinlauKL: Thanks for catching this.

Multiset/Multiset.playground/Sources/Multiset.swift
simonwhitaker: 😲 😲 😍

kelvinlauKL:
:+1:

PR #625, opened 2017-09-26
update to radix sort

jakecast:
### Checklist

### Description

tracking the largest value in the array on the first iteration of the bucket sort allows you to do one less loop and drops the need for a `done` variable.
also updated the bucket initialization to use the `(repeatingElement: count:)` initializer and switched to using an iterator for looping through the buckets instead of accessing them based on the range loop.

### DIFF
--- a/Radix Sort/radixSort.swift
+++ b/Radix Sort/radixSort.swift
@@ -5,43 +5,35 @@
 */
 
 
-func radixSort(inout arr: [Int] ) {
+func radixSort(arr: inout [Int]) {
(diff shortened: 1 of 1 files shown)


kelvinlauKL:
Thank you! Welcome to the club 👍

jakecast:
I pushed a very small change to set the number of buckets to be initialized using the radix property. not a big deal given the input is constrained to an Int array, but worth changing if the sort was updated to support different base sizes.

PR #629, opened 2017-10-03
Depth first search

JulioBBL:
### Checklist

### Description

 

The class Graph, is a representation of a Digraph, which is a Graph with oriented edges, a normal Graph would have two way edges


vincentngo:
@JulioBBL It seems like you just changed the name from **Graph** to **Digraph**, i'm not sure if I see any real benefit with this. Could you please elaborate?

remlostime:
I think graph could be two ways or one way. `Digraph` is kind of hard to understand what's `Di` mean?

vincentngo:
@JuIioCesar closing this since I don't think changing just the name to Digraph makes any improvements to the implementation.

JuIioCesar:
@vincentngo Wrong Julio cc: @JulioBBL

PR #685, opened 2018-01-06
Improvements to Heap

DXXL:
### Checklist

### Description

 

- renamed "elements" to "nodes" to clarify the difference between array and heap. This also unifies code and comments, which were sometimes referring to the array as "nodes".
- made nodes publicly readable but only settable within the heap itself. To change the nodes/operate on the heap, the provided functions should be used, to guarantee the heaps structure.
- renamed isOrderedBefore to orderCriteria. The variable name isOrderedBefore did not clearly explain what the variable is for/what the closure stored within does
- added method to get node at a specific index
- replace method should not force the replaced value to be higher/lower than its parent. Instead replace should remove the node to be replaced and insert the given value as a new one in order to preserve the heap's structure. Asserting crashes the whole program.
- added @discardableResult to removeAt. If remove is @discardableResult removeAt should be as well in for consistency.
- replaced while true in shiftDown with a recursive call. while true is bad programming style.
- simplified indexOf method
- added a removeNode method for heaps with equatable data type
- improved, adjusted and added descriptions
- made functions private where fileprivate was not needed
- added tests for new functions and tested everything
- updated README where necessary (to conform to new code, only a few changes)


Heap/Heap.swift
```diff
@@ -4,58 +4,55 @@
 //
 
 public struct Heap<T> {
+  
   /** The array that stores the heap's nodes. */
-  var elements = [T]()
+  private(set) var nodes = [T]()
```
kelvinlauKL: Let's declare this as simply `private` instead. The internal storage of the heap shouldn't be something people need to inspect.

Heap/Heap.swift
```diff
@@ -4,58 +4,55 @@
...
   /**
-   * Converts an array to a max-heap or min-heap in a bottom-up manner.
+   * Creates the max-heap or min-heap from an array, in a bottom-up manner.
    * Performance: This runs pretty much in O(n).
    */
-  fileprivate mutating func buildHeap(fromArray array: [T]) {
-    elements = array
-    for i in stride(from: (elements.count/2 - 1), through: 0, by: -1) {
-      shiftDown(i, heapSize: elements.count)
+  private mutating func buildHeap(fromArray array: [T]) {
```
kelvinlauKL: This signature could also use some touch up.

I suggest `buildHeap(from array: [T])`. Also thought of changing the name to `configureHeap(from array: [T])`.

The statement "buildHeap" suggests we're creating a new heap, but what is really going on is we're assigning an array to `nodes`, and structuring it as a heap.

Heap/Heap.swift
```diff
@@ -89,69 +86,79 @@ public struct Heap<T> {
    * value (for a min-heap).
    */
   public func peek() -> T? {
-    return elements.first
+    return nodes.first
+  }
+  
+  /** Returns the node at given index */
+  public func node(at i: Int) -> T? {
```
kelvinlauKL: Do you have a use case in mind for this method (besides using it internally for testing)? I don't see a use case for retrieving a value at the middle of a heap.

Heap/Heap.swift
```diff
@@ -89,69 +86,79 @@ public struct Heap<T> {
...
-    } else if elements.count == 1 {
-      return elements.removeLast()
-    } else {
-      // Use the last node to replace the first one, then fix the heap by
-      // shifting this new first node into its proper position.
-      let value = elements[0]
-      elements[0] = elements.removeLast()
-      shiftDown()
-      return value
+    if !nodes.isEmpty {
```
kelvinlauKL: Instead of this `if` statement, let's swap it out with `guard`:

```
guard !nodes.isEmpty else { return nil }

// rest of the code
```

This avoids a level of pyramiding, and leverages the compiler aided exit through the `else` statement from `guard`.

Heap/Heap.swift
```diff
@@ -89,69 +86,79 @@ public struct Heap<T> {
...
   
   /**
-   * Removes an arbitrary node from the heap. Performance: O(log n). You need
-   * to know the node's index, which may actually take O(n) steps to find.
+   * Removes an arbitrary node from the heap. Performance: O(log n).
+   * Note that you need to know the node's index.
    */
-  public mutating func removeAt(_ index: Int) -> T? {
-    guard index < elements.count else { return nil }
+  @discardableResult public mutating func removeAt(_ index: Int) -> T? {
```
kelvinlauKL: Let's update this signature to `remove(at index: Int)`.

hollance:
I don't mind most of these changes (though I din't necessarily think thar using the term nodes is better than elements, for example) but saying that a recursive function is better than a loop is simply not true. It might make the code smaller (but only more readable if you understand recursion well) but it also introduces the possibily of stack overflow, something the loop does not. (And is possibly slower due to the extra function calls.)

Heap/Heap.swift
DXXL: No, I really only used it for testing purposes

Heap/Heap.swift
DXXL: Thanks, for your further improvements really like them.

DXXL:
I just read the code and thought about some things I would improve. Your right that it is unimportant if we call it 'nodes' or 'elements'. But the comments sometimes said 'nodes' and sometimes 'elements' and I wanted to unify this, so I had to decide between 'nodes' and 'elements'. You could as well change the comments.
You can, of course, use a 'while true' loop again, but this has its own downsides. I think it's just a question of style, so actually, it does not matter to me what is used.
Thank you for your reply!

hollance:
Apologies if I sounded a little snippy, I've got a stomach bug today. It's just that I can defend many of the choices I made with good arguments, and it seemed a bit like you were saying these were bad choices. But equally good arguments can be made for your changes. :-)

Heap/Heap.swift
kelvinlauKL: I suggest taking this out then. Let's think of another way to test without exposing this member function.

kelvinlauKL:
👍 

@DXXL There's a build error lurking somewhere, could you fix that and apply the other suggested changes?

DXXL:
Yes, I'm about to adopt your suggestions. The error is because HeapSort accesses the elements, I think. I will have a look.

kelvinlauKL:
Looks good. Thanks @DXXL!

PR #697, opened 2018-03-12
Refactors Binary Search implementation to be more Swifty.

philium:
### Checklist

### Description

This involved converting the global function into methods of
`Collection`. By doing so, the recursive version no longer needs to
take a search range as a parameter. It also made both the recursive and
iterative versions of the algorithm usable on other collection types
besides `Array`. Tests were also refactored to test both versions of
the algorithm.


kelvinlauKL:
Hey @philium, sorry for the late response on this PR. This PR was very comprehensive and I regret not addressing it promptly.

That said, I've also considered extending binary search as a method inside `Collection`, but I encountered a dilemma: 

1. Binary search only works on **sorted** collections, and putting this function inside `Collection` sets the precedent for non-general purpose algorithms to pollute an otherwise general purpose collection of functions.

2. Again, binary search only works on **sorted** collections. Allowing binary search to work with all types conforming to `Collection` is not desirable, because not all `Collection` conformers have the concept of order. Dictionaries are a good example of a type that conforms to `Collection` but cannot be sorted. Binary search becomes an erroneous API for those collections that cannot be sorted. 

For those reasons, I don't think adding binary search as an extension of `Collection` is the right move.

philium:
@kelvinlauKL Thank you very much for reviewing my PR. I appreciate your comments. Do you mind if I respond to them?

> putting this function inside Collection sets the precedent for non-general purpose algorithms to pollute an otherwise general purpose collection of functions.

As you stated, binary search only works on sorted collections. But an array is not inherently sorted. It may be sorted or it may be unsorted. Unfortunately we currently have no way to express a sorted collection with Swift's type system. Yet we still use binary search with arrays. Why? Because they are an ordered collection that *can* be sorted. By generalizing binary search to collections whose elements are comparable, it becomes useful on other collection types that *can* be sorted, besides `Array`:

- `ContiguousArray`
- `CollectionOfOne`
- `EmptyCollection`
- `AnyCollection`
- `FlattenCollection`
- `LazyCollection`
- `LazyDropWhileCollection`
- `LazyFilterCollection`
- `LazyMapCollection`
- `LazyPrefixWhileCollection`
- `Slice`
- `ArraySlice`
- Any custom type conforming to `Collection` whose `Element` is comparable.

> Allowing binary search to work with all types conforming to Collection is not desirable, because not all Collection conformers have the concept of order. Dictionaries are a good example of a type that conforms to Collection but cannot be sorted.

Since the algorithm is constrained to collections whose elements are comparable, the algorithm doesn't work with `Dictionary`. Swift's dictionary element type is a tuple and tuples are not comparable. A better example would have been `Set`. Swift's set element can be comparable, so the algorithm is exposed on sets whose element is comparable.

Which is worse? Exposing the algorithm on a select few collection types that cannot be sorted or preventing the algorithm from being used on over a dozen other collection types whose elements could be sorted? I would argue the later.

Could I please ask you to reconsider your decision?

hollance:
It was my goal for this project to illustrate how such algorithms work, not necessarily make the best possible implementation. So these articles should be written to explain the algorithm first and foremost. Optimizations and improvements are good to include but are best placed at the bottom of the article, after the algorithm has been explained in very straightforward code.

In this case it seems like a good idea to use Collection instead of Array but personally I would first explain the Array version since most developers will be familiar with arrays, and then show how to make it more suitable for other collections too and why this is a good idea (not everyone will be familiar with all these protocols). 

That way we only show one concept at a time: first the idea of binary search (using something most developers will be familiar with), then how to integrate it into the Collection stuff.

philium:
@hollance Thank you for your comments. Do you mind if I ask you a couple questions?

> these articles should be written to explain the algorithms first and foremost.

Do you have a specific example from the readme where the explanation of the algorithm is not as straightforward as a result of implementing it on Collection?

> I would first explain the Array version since most developers will be familiar with arrays…(not everyone will be familiar with all these protocols).

It was a conscious decision to use Swift as the language for this repository. Is the target audience Swift developers, non-Swift developers, or both? Protocols are a part of developing in Swift. Any Swift developer should be familiar with the Collection collection of protocols as they are foundational to the language. They are also fundamental to how arrays work. While an array-based implementation might make more sense for a non-Swift developer, to a Swift developer, it feels out of place as of Swift 2.

I love the idea of progressive disclosure, showing one concept at a time. And I think I’ve done that by sticking with Array in the examples in the readme. 

Have you had a chance to watch WWDC 2018 Session 223, “Embracing Algorithms”? I just watched it a couple weeks ago. The approach I took seems to fit with Apple’s advice to make algorithms generic and reusable. I’m curious if you have any thoughts on that.

Thank you very much for your time.

hollance:
@philium I don't believe most Swift developers are intimately familiar with the family of Collection protocols. I certainly am not. I do believe your changes make the code better but I don't think explaining everything in terms of Collection right away is a good idea. But I'm no longer a maintainer for this repo so I'll leave that up to the other guys. Just my 2 cents. :-)

hollance:
This discussion would be useful to add to the explanation of binary search. ;-)

philium:
@kelvinlauKL I could certainly do that. Are you referring to updating just the readme or the associated project and playground as well? I think the readme could be be updated to start with an array-based implementation and then transition to a collection-based implementation. Since the latter is more useful, I would leave the Swift files the way they are.

kelvinlauKL:
I'd suggest updating both the readme and associated project/playground to contain both implementations.

kelvinlauKL:
Closing this due to inactivity. Feel free to re-open if you're still working on this!

PR #703, opened 2018-03-30
Refactors LinkedList implementation to be more Swifty.

SeRG1k17:
List of changes:

- Removed several assert and force unwrapping.
- Reworked property var head: Node?
- Removed unnecessary self.
- Default for replaced to forEach


SeRG1k17:
How does this conflict?

PR #705, opened 2018-03-31
Refactors LinkedList implementation to be more Swifty.

SeRG1k17:
List of changes:

- Removed several assert and force unwrapping.
- Reworked property var head: Node?
- Removed unnecessary self.
- Default for replaced to forEach


vincentngo:
@SeRG1k17 Sorry for the delay, we are picking things up now. I will be reviewing this.

SeRG1k17:
@vincentngo Okay 😃

PR #712, opened 2018-04-30
[Binary Search Tree] [enum] Fix height

ruslanskorb:
### Checklist

### Description

In accordance with the [terminology](https://en.wikipedia.org/wiki/Tree_(data_structure)#Terminology), leaf nodes have height **zero**, and a tree with only a single node (hence both a root and leaf) has height **zero**. Conventionally, an empty tree (tree with no nodes, if such are allowed) has height **−1**.

### DIFF
--- a/Binary Search Tree/README.markdown
+++ b/Binary Search Tree/README.markdown
@@ -615,8 +615,8 @@ This implementation is recursive, and each case of the enum will be treated diff
 
   public var height: Int {
     switch self {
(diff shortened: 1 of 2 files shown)


ruslanskorb:
@vincentngo ping (updated)

ruslanskorb:
@kelvinlauKL Hi, could you take a look at this PR? Thanks!

PR #730, opened 2018-07-20
fix delete

strohtennis:
### Checklist

### Description
By having a BST that is defined by using recursive BSTs you limit the ability to have an empty tree.
The current code has an issue not only with the empty tree, but also with deleting the root node.  Currently, if you delete the root node, the entire tree is deleted except for the root.
By using a separate BSTNode class, the tree itself can have a reference to the root node and do all operations from there.  Some of the variables and methods got moved to the BSTNode, although it may be better to reference some of the methods from the tree using the root.
The parent node instance of the BSTNode isn't needed for delete.  It was left in for some of the other methods that use it.
The corresponding other playgrounds that reference the BST along with the tutorial will need to be updated as well.  Please let me know if you would like me to work on that as part of this PR.
Cheers, Eric

### DIFF
--- a/Binary Search Tree/Solution 1/BinarySearchTree.playground/Sources/BinarySearchTree.swift
+++ b/Binary Search Tree/Solution 1/BinarySearchTree.playground/Sources/BinarySearchTree.swift
@@ -1,346 +1,422 @@
 /*
-  A binary search tree.
-
-  Each node stores a value and two children. The left child contains a smaller
-  value; the right a larger (or equal) value.
-
-  This tree allows duplicate elements.
-
-  This tree does not automatically balance itself. To make sure it is balanced,
-  you should insert new values in randomized order, not in sorted order.
-*/
-public class BinarySearchTree<T: Comparable> {
-  fileprivate(set) public var value: T
-  fileprivate(set) public var parent: BinarySearchTree?
-  fileprivate(set) public var left: BinarySearchTree?
-  fileprivate(set) public var right: BinarySearchTree?
-
-  public init(value: T) {
-    self.value = value
-  }
-
-  public convenience init(array: [T]) {
-    precondition(array.count > 0)
-    self.init(value: array.first!)
-    for v in array.dropFirst() {
-      insert(value: v)
+ Each node stores a value and two children. The left child contains a smaller
+ value; the right a larger (or equal) value.
+ */
+public class BSTNode<T: Comparable> {
+    var value: T
+    var left: BSTNode?
+    var right: BSTNode?
+    var parent: BSTNode?
+    
+    init(initialValue: T, leftNode: BSTNode? = nil, rightNode: BSTNode? = nil, parentNode: BSTNode? = nil) {
+        value = initialValue
+        left = leftNode
+        right = rightNode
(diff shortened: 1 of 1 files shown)


kelvinlauKL:
Hey @strohtennis, thanks for the PR! I'm also a fan of having a manager class that manages my nodes. More flexibility and a cleaner interface, what's there to lose! 👍

However, I have a few concerns:

> The current code has an issue not only with the empty tree, but also with deleting the root node. Currently, if you delete the root node, the entire tree is deleted except for the root.

The current implementation is designed for you to manage the nodes directly (instead of having a separate class to manage the nodes). Deleting the root node should be synonymous to setting the reference of the root node that you're holding to `nil`. Could you elaborate on the situation where deleting the root node wouldn't actually delete the node?

Another concern I have is the scope of this refactor. There's a few things that needs to be done:

1. Update tree implementation.
2. Update tests based on the new implementation.
3. Update the README file to address the new implementation.

I don't mind reviewing a rewrite, but please be advised that the bar would be pretty high to replace the current writeup!

strohtennis:
> "The current implementation is designed for you manage the nodes directly"

A binary search tree class is what should be managing the nodes.  That's the convenience of the BST.   I would think that you would want your class to be named TreeNode or something similar because it does require a lot of implementation outside of the class for the BST to be managed correctly.

It would be convenient if I could call `delete(value: T)` but that isn't possible with this implementation.

To delete a value you have to handle searching to find the value first, and then call remove on the node itself.

1.  `let arr = [ 47, 22, 76, 12, 33, 67, 89, 5,13,28,37, 79,91,52, 64]`
2.  `var tree = BinarySearchTree(array: arr)`

3.  `let node47 = tree.search(value: 47)`
4. ` node47?.remove() `

Notice that your BST now contains only one BST with nil left & right and a value of 47.  Instead of line 3 & 4 above you would have to do more managing yourself to handle deleting the root and the possibility of an empty tree.

`if let node47 = tree.search(value: 47) {
    let isRoot = node47.isRoot
    let newTree = node47.remove() ?? BinarySearchTree(array: [])
    if isRoot {
        tree = newTree
    }
}`

Do you really want to do all of that work every time you want to delete a value from a tree?  That's why that implementation should be in the BST class itself.  

What would concern me the most is the memory location of the tree itself.  Once you remove the root from the tree, a dangerous situation could arise that the pointer to the original tree root may not get updated in other features of your code, possibly losing all the data that should still be in the tree.  

If this repository is intended to be used to help developers, it should have high standards of best practices.  I understand that the scope for this change is significant.  I have the time to address those if you agree that it is warranted.  I don't want to spend the time if you feel that the current implementation is sufficient.

kelvinlauKL:
Good points.

I pondered some more and I think a refactor would be good. In addition to the points you've brought up, the current implementation couples the tree logic with the node. I don't feel that is quite correct, since the same node class can be shared in other forms of binary trees, such as AVL trees. 

It feels much nicer to concentrate the tree characteristics inside a class managing the nodes, rather than inside the nodes directly. 

Thank you for bringing this up. If you're up for it, please go ahead with the refactor.

Issue #731, opened 2018-07-25
Simulated Annealing demo crashes


pstoehr:
### Brief Intro

Simulated Annealing demo crashes

### More Details

Crash in code line
    swap(&self.tour[i], &self.tour[j])
with error stack 
Simultaneous accesses to 0x103c053c0, but modification requires exclusive access.
Previous access (a modification) started at Test2`Tour.shuffle() + 965 (0x1000033b5).
Current access (a modification) started at:
0    Test2                              0x0000000100471360 swift_beginAccess + 511
1    Test2                              0x0000000100002740 Tour.tour.materializeForSet + 34
2    Test2                              0x0000000100002ff0 Tour.shuffle() + 1115
3    Test2                              0x0000000100002fd0 protocol witness for SAObject.shuffle() in conformance Tour + 16
4    Test2                              0x00000001000037f0 SimulatedAnnealing<A>(initial:temperature:coolingRate:acceptance:) + 327
5    Test2                              0x0000000100001810 main + 853
6    libdyld.dylib                      0x00007fff71dceed4 

System:
Version 9.4.1 (9F2000)

pstoehr:
Same problem with Xcode Version 10.0 beta 3 (10L201y)

kelvinlauKL:
Thank you for brining this up @pstoehr!

dynamite8:
@pstoehr - I attempted to fix this issue and resolved all the build issues except that I can't test it. Would you be able to try out the changes in my branch?  I fixed it based on Xcode 10 beta version though.

pstoehr:
Have you already committed the new sources?

dynamite8:
Click on the link #733 and you'll see the two files that I updated @pstoehr.

kelvinlauKL:
The particular error "Simultaneous accesses to <memory address>" suggests it has something to do with the update that happened last year on exclusive memory access: 
https://devstreaming-cdn.apple.com/videos/wwdc/2017/402ynph39nk5sn4222/402/hls_vod_mvp.m3u8

(46 min mark)

PR #733, opened 2018-08-03
Fixes Issue#371: Simulated Annealing demo crashes

dynamite8:
\### Checklist

- [ X] I've looked at the [contribution guidelines](https://github.com/raywenderlich/swift-algorithm-club/blob/master/.github/CONTRIBUTING.md).
- [ ] This pull request is complete and ready for review.

### Description

Updated code based on Xcode 10 beta version.  This is not the current version supported by SAC but in case we move towards Xcode 10 in the future, this is here.

Asking the original issue opener to review this.

### DIFF
--- a/Simulated annealing/simann.swift
+++ b/Simulated annealing/simann.swift
@@ -22,16 +22,6 @@
   import Glibc
 #endif
 
-public extension Double {  
-  public static func random(_ lower: Double, _ upper: Double) -> Double {
-    #if os(OSX)
-      return (Double(arc4random()) / 0xFFFFFFFF) * (upper - lower) + lower
-    #elseif os(Linux)
-      return (Double(random()) / 0xFFFFFFFF) * (upper - lower) + lower
-    #endif
-  }
-}
-
 protocol Clonable {
   init(current: Self)
 }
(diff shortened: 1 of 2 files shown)


pstoehr:
From my point of view the bug is still there!
As mentioned in the online documentation, the swap() routine should not be used for mutable collections. If two elements of mutable collection should be swapped, swapAt() have to be used!

dynamite8:
@pstoehr - there were compilation errors so those are fixed in this PR now.  How are you running this?  Can you pass the details?  If it's only needed to change the methods, then that should be easy.

kelvinlauKL:
These changes look good with respect to Xcode 10.0. 

I think the author of this algorithm is running on Linux, thus there wasn't an Xcode playground accompanying this. 

@dynamite8 would you be interested in authoring a playground for this topic?

kelvinlauKL:
Merging this for Xcode 10.0 compatibility. @pstoehr I'll leave the #731 issue open for further discussion.

pstoehr:
Come on folks, It seems that you've included a **new bug**!
The method shuffle() of the extension of the class Tour calls itself with an endless recursion!

To find this bug I just downloaded the file simann_example.swift, created a new Xcode project using Version 10.0 beta 5. Than 
    Compile -> Run -> Crash

By the way, the compiler gives the warning "All paths through tis function will call itself" and this is a strong hint that you have an endless recursion!

From my point of view the implementation should look like 
```
    func shuffle() {
        tour.shuffle()
    }
```
Would you please be so kind to check whether this really fixes the bug!

PR #741, opened 2018-09-17
optimize special case return early

neonio:
### Checklist


kelvinlauKL:
Hey @neonio!

This special case makes sense and works. However, I feel this is extra cruft for a very specific case that doesn't happen all too often. I'd like to keep the implementation simpler, so newer readers don't have to digest this extra bit of code to understand how mergesort works. 

Best,
Kelvin

Issue #748, opened 2018-10-03
Swift 4.2 Migration


kelvinlauKL:
### Brief Intro

Every year, we go through the repository to make sure all the `README` and playgrounds are up-to-date with the latest Xcode. 

## Migration Procedure:

1. Choose a topic below that hasn't been checked off
2. Check the pull requests to see if anyone is working on your chosen topic
3. Make a pull request to let others know that you're currently working on your chosen topic
4. 

**Remove** code snippet at the top of each playground file:

```
// last checked with Xcode 9.0b4
#if swift(>=4.0)
   print("Hello, Swift 4!")
#endif
```

* Make sure the code for the topic compiles correctly with the latest Xcode 10. 
* Make sure that the README.md file reflects the updated playground

## Current Progress

- [x] 3Sum and 4Sum
- [x] AVL Tree
- [ ] All-Pairs Shortest Paths
- [x] Array2D
- [x] B-Tree
- [x] Binary Search Tree
- [x] Binary Search
- [x] Binary Tree
- [ ] Bit Set
- [x] Bloom Filter
- [x] Bounded Priority Queue
- [x] Boyer-Moore-Horspool
- [x] Breadth-First Search
- [ ] Brute-Force String Search
- [ ] Bubble Sort
- [x] Bucket Sort
- [ ] Comb Sort
- [ ] Combinatorics
- [x] Convex Hull
- [x] Count Occurrences
- [x] Counting Sort
- [x] Depth-First Search
- [x] Deque
- [x] Dijkstra Algorithm
- [ ] DiningPhilosophers
- [x] Egg Drop Problem
- [ ] Encode and Decode Tree
- [x] Fixed Size Array
- [x] Fizz Buzz
- [x] GCD
- [x] Graph
- [x] Hash Set
- [x] Hash Table
- [x] HaversineDistance
- [x] Heap Sort
- [x] Heap
- [x] Huffman Coding
- [ ] Insertion Sort
- [x] K-Means
- [x] Karatsuba Multiplication
- [x] Knuth-Morris-Pratt
- [x] Kth Largest Element
- [x] LRU Cache
- [x] Linear Regression
- [x] Linear Search
- [x] Linked List
- [ ] Longest Common Subsequence
- [x] Merge Sort
- [x] Miller-Rabin Primality Test
- [x] Minimum Edit Distance
- [x] Minimum Spanning Tree (Unweighted)
- [x] Minimum Spanning Tree
[...]

PR #759, opened 2018-10-04
[Swift 4.2] Bloom Filter

SpacyRicochet:
### Checklist


SpacyRicochet:
References #748.

Bloom Filter/README.markdown
```diff
@@ -100,4 +100,29 @@ public func query(_ value: T) -> Bool {
 
 If you're coming from another imperative language, you might notice the unusual syntax in the `exists` assignment. Swift makes use of functional paradigms when it makes code more consise and readable, and in this case `reduce` is a much more consise way to check if all the required bits are `true` than a `for` loop.
 
-*Written for Swift Algorithm Club by Jamil Dhanani. Edited by Matthijs Hollemans.*
+## Another approach
+
+Another approach to create different hashes of an element for use in the Bloom filter, is to use the same hash function for every iteration, but combine it with different random numbers. This can help, because finding good hashing functions is hard, but combining them is equally non-trivial.
```
kelvinlauKL: This part is unclear for me. 

My understanding of hash functions is that it should be deterministic -- "hello world!" should hash to the same value during the lifetime of a single program. 

If you combine the result of the hash function with different random numbers, wouldn't that just be akin to producing random numbers?

SpacyRicochet: For the Bloom filter, the actual hash isn't important, as long as it is consistent and a proper hash. So if you guarantee that you acquire the hash for the same object in the same way during the lifetime of the Bloom filter, you're fine.

This approach would create random numbers at the initialisation of the Bloom filter, and use them consistently as modifiers on an object's hash, effectively creating a new hash function per random number. So the hash would still be same for object A during the lifetime of that Bloom filter.

Make sense? The linked blog post about it is probably clearer.

kelvinlauKL: Ah, I see what you're getting at. I read this as "everything time you use the hash function, combine it with a random number."

I propose we change this paragraph to the following:


>In the previous section, you learnt about how using multiple different hash functions can help reduce the chance of collisions in the bloom filter. However, good hash functions are difficult to design. A simple alternative to multiple hash functions is to use a set of random numbers.

>As an example, let's say a bloom filter wants to hash each element 15 times during insertion. Instead of using 15 different hash functions, you can rely on just one hash function. The hash value can then be offset by 15 different values to form the indices for flipping. This bloom filter would initialize a set of 15 random numbers ahead of time and use these values during each insertion.

Let me know if that sounds fine with you. If this is okay, I'll go ahead and make the change to merge it in.

SpacyRicochet: Updated the README.

kelvinlauKL:
Thanks @SpacyRicochet!

PR #829, opened 2018-11-01
Counting Sort working with Swift 4.2

iillx:


richard-ash:
Thanks for the update! For reference this migrates Counting Sort in #748

PR #831, opened 2018-11-01
Shunting Yard working fine with Swift 4.2 and Xcode 10

iillx:


richard-ash:
For issue #748. Thanks for the update!

PR #832, opened 2018-11-01
Updated random number generation to Swift 4.2

iillx:


richard-ash:
Thanks for the update. For reference #748

Issue #717, opened 2018-05-21
Boyer-Moore-Horspool algorithm implementation is not correctly


luoqisheng:
### Brief Intro

```
if c == lastChar {
    if k = backwords() { return k }
    ///  skipTable[c]  == skipTable[lastChar] == 0
    i = max(skipTable[c] ?? patternLength, 1)
    .....
}
```

PR #835, opened 2018-11-06
Fix deprecated methods on Boyer-Moore-Horspool 

pakosaldanaort:
I fixed some deprecated method and I updated them to swift 4.2, I changed the return type Index for Int. This solves the issue #717


richard-ash:
Thanks!

PR #849, opened 2018-12-16
Fixed error in description

amichnia:
### Checklist

- [√] I've looked at the [contribution guidelines](https://github.com/raywenderlich/swift-algorithm-club/blob/master/.github/CONTRIBUTING.md).
- [√] This pull request is complete and ready for review.

### Description

 

There is an error in Comb Sort description, the name of O(n^2) should be polynomial, not expotential. This PR fixes that, also adding naming for o(n logn) -- quasilinear

### DIFF
--- a/Comb Sort/README.markdown
+++ b/Comb Sort/README.markdown
@@ -64,8 +64,8 @@ This will sort the values of the array into ascending order -- increasing in val
 ## Performance
 
 Comb Sort was created to improve upon the worst case time complexity of Bubble Sort. With Comb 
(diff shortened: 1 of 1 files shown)


richard-ash:
Great catch thanks 😃

One thought. `quasilinear` seems like a fairly complex term and may confuse readers, it may make more sense to just leave the `O(nlogn)` with no name.

amichnia:
What about log-linear? It more or less matches what's in O(...) so should seem less complex.

richard-ash:
Sure! Thanks again for the update

PR #855, opened 2019-01-31
Implement bit shift operations in BitSet

Heraldk:
### Checklist


Bit Set/BitSet.swift
```diff
@@ -221,6 +221,51 @@ prefix public func ~ (rhs: BitSet) -> BitSet {
...
 
+// MARK: - Bit shift operations
+
+/*
+ Note: For bitshift operations, the assumption is that any bits that are
+ shifted off the end of the end of the declared size are not still set.
+ In other words, we are maintaining the original number of bits.
+ */
+
+public func << (lhs: BitSet, numBitsLeft: Int) -> BitSet {
```
richard-ash: Looks like you already implemented this method above. Is there a reason for the duplication?

Bit Set/BitSet.swift
```diff
@@ -221,6 +221,51 @@ prefix public func ~ (rhs: BitSet) -> BitSet {
...
+        if (i - offset - 1 >= 0) {
+            out.words[i] |= lhs.words[i - offset - 1] >> (lhs.N - shift)
+        }
+    }
+    
+    out.clearUnusedBits()
+    return out
+}
+
+public func >> (lhs: BitSet, numBitsRight: Int) -> BitSet {
```
richard-ash: Same here

Bit Set/BitSet.swift
Heraldk: For both of these, I'm updating the source that's in the playground as well as the main source file. I couldn't find an obvious way these two files were linked (if they are at all) so it seemed like I needed to update both source files to match.
Unless I'm missing something about how playgrounds work?

Heraldk:
> Code looks good 👍 Left one minor comment inline. Would you mind adding a section in the README explaining these two methods?
> 
> Thanks again!

Sure, I'll add some README notes soon. Might be a day or two :)

Bit Set/BitSet.swift
richard-ash: Hmm yea, that's a good point. Would you mind deleting the entire file? (`Bit Set/BitSet.swift`) Ideally, we'd want only a single set of files with all the source code

richard-ash:
> > Code looks good 👍 Left one minor comment inline. Would you mind adding a section in the README explaining these two methods?
> > Thanks again!
> 
> Sure, I'll add some README notes soon. Might be a day or two :)

Sounds good! Looking forward to reading it 🙂

Bit Set/BitSet.swift
Heraldk: Is that the new pattern? There are a lot of instances in this repository where the code is duplicated in the playground either as an inclusion or just the code copy-pasted into the playground source file.

I can remove the file outside the playground if you like :)

Heraldk:
I added a README section for bit-shift operations, and removed the duplicate source file as requested.

Bit Set/BitSet.swift
richard-ash: Yea was looking around an noticed this too. I'll start pointing it out more as people update the repo. Even if the code is copy-pasted its a super easy thing to forget to update. Much easier to just have one source of truth.

That'd be great thanks :)

richard-ash:
> I added a README section for bit-shift operations, and removed the duplicate source file as requested.

Thanks, 😄the section looks good, nice work!

Heraldk:
Anything left on me that prevents this from being merged? No rush - just making sure I'm not missing something :)

richard-ash:
Oops thanks for the reminder

PR #856, opened 2019-02-01
reserveCapacity missing in README for MergeSort

amakukha:
### Checklist

### Description

 

reserveCapacity was missing in README for Merge Sort despite being present in the sample code. 

In my tests, reserving capacity unconditionally right after creating array for merging worked best, sorting a shuffled array with 1 million integers in 11.6% less time.  

Even though it doesn't reduce time complexity, I think that a simple line of code that has so much effect on efficiency should not be omitted. And removing reserveCapacity when required capacity is known precisely in advance is certainly not very _Swift_'y.


richard-ash:
Cool 💯 thanks for the optimization!!

PR #868, opened 2019-03-16
Add getPredecessor() and getKey() methods to RBTreeNode.

dcherednikov:
### Checklist

### Description

 

Implement getPredecessor() method for RBTreeNode.
Implement getKey() method for RBTreeNode. 
Update docs.

### DIFF
--- a/Red-Black Tree/README.markdown
+++ b/Red-Black Tree/README.markdown
@@ -29,6 +29,7 @@ From [CLRS]
 ## Methods
 
 Nodes:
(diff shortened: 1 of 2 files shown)


dcherednikov:
Just in case if some explanations are needed as to why this PR is created.
The docs originally stated that the get predecessor operation has a logarithmic worst case performance but no such method was actually implemented. This operation is required for some algorithms that make use of Red-Black trees, for instance Shamos-Hoey algorithm. And the getKey() method makes a node actually usable outside the RedBlackTree implementation.

PR #869, opened 2019-03-20
Update Shuffle

passerbyloo:
according to https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle

`checks if j ≠ i may be omitted`, and also remove checking logic will optimise performance(tested by 100000 times random).

### Checklist

### DIFF
--- a/Shuffle/README.markdown
+++ b/Shuffle/README.markdown
@@ -96,6 +96,8 @@ public func shuffledArray(_ n: Int) -> [Int] {
   var a = [Int](repeating: 0, count: n)
   for i in 0..<n {
     let j = Int.random(in: 0...i)
+    // for the Fisher–Yates_shuffle's pseudo code implement in wiki, it will check if i != j
+    a[i] = a[j]
     a[j] = i
   }
   return a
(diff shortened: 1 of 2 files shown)


Shuffle/README.markdown
```diff
@@ -96,9 +96,6 @@ public func shuffledArray(_ n: Int) -> [Int] {
   var a = [Int](repeating: 0, count: n)
   for i in 0..<n {
     let j = Int.random(in: 0...i)
-    if i != j {
-      a[i] = a[j]
```
richard-ash: Removing this line breaks the algorithm. You're right the check `if i != j` is unnecessary however you still need the line `a[i] = a[j]`

Shuffle/Shuffle.swift
```diff
@@ -30,9 +30,6 @@ public func shuffledArray(_ n: Int) -> [Int] {
   var a = Array(repeating: 0, count: n)
   for i in 0..<n {
     let j = random(i + 1)
-    if i != j {
-      a[i] = a[j]
```
richard-ash: Same here

Shuffle/README.markdown
passerbyloo: updated

richard-ash:
Cool, thanks for the update 🙂

PR #875, opened 2019-04-04
Brute Force String Search is updated as "String.characters" is deprecated

mohankrishnameruva:
### Checklist

### Description

 
The current Brute-force String search algorithm is using string.characters method which is deprecated. I have updated the algorithm without changing any functionality.

### DIFF
--- a/Brute-Force String Search/README.markdown
+++ b/Brute-Force String Search/README.markdown
@@ -19,33 +19,27 @@ let animals = "🐶🐔🐷🐮🐱"
 animals.indexOf("🐮")
 
 // Output:
-<String.Index?> 6
+<String.Index?> 3
 ```
-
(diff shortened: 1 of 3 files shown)


mohankrishnameruva:
Thanks for the review, I have updated the code as per your comments could you please look at it.

Brute-Force String Search/README.markdown
```diff
@@ -19,30 +19,26 @@ let animals = "🐶🐔🐷🐮🐱"
 animals.indexOf("🐮")
 
 // Output:
-<String.Index?> 6
+<String.Index?> 3
 ```
 
-> **Note:** The index of the cow is 6, not 3 as you might expect, because the string uses more storage per character for emoji. The actual value of the `String.Index` is not so important, just that it points at the right character in the string.
```
kelvinlauKL, line 25: Are you certain this behavior has changed?

PR #890, opened 2019-09-03
insertion sort: performance improvements

nickm01:
### Checklist

### Description
Having gone through insertion sort, I noticed that the original algorithm did not perform very well with large numbers of numbers.  The reason being - it "shuffles" the numbers to the left one place at a time.  Not sure that's a requirement of the algorithm.  Better to do one "move" when the right spot is found.  This is 2X quicker in my tests for 1000+ results.
Here's a good example of what's going on in this variant of the algorithm (you'll see how 3 leaps over 5 and 6 rather than gets placed between 5 and 6 first).
https://en.wikipedia.org/wiki/Insertion_sort#/media/File:Insertion-sort-example-300px.gif.

Also included better naming, removed duplicate code and updated the comments.

### DIFF
--- a/Insertion Sort/InsertionSort.playground/Contents.swift
+++ b/Insertion Sort/InsertionSort.playground/Contents.swift
@@ -1,43 +1,40 @@
-//: Playground - noun: a place where people can play
-
 /// Performs the Insertion sort algorithm to a given array
 ///
 /// - Parameters:
 ///   - array: the array of elements to be sorted
-///   - isOrderedBefore: returns true if the elements provided are in the corect order
+///   - isOrderedBefore: function that given to inputs returns whether should be order before or after.  Allows for different sorts (e.g. change in direction).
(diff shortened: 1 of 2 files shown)


richard-ash:
Cool optimization, thanks for the Pull Request 😃

Given that the purpose of the piece is explanatory rather than efficiency I'd rather we leave the algorithm in its basic, naive form.

However, I still think this optimized `insertionSort` is worth including! Can you create a second function `insertionSortOptimized` and place your code there? 

Also would you be up for updating the README? It'd be fantastic if you could add a short section on the optimization you made and how it's faster than the original.

Thoughts, how's that sound?

nickm01:
Thanks @richard-ash!  As @hollance pointed out, I was testing performance in playgrounds which isn't a good test.  Actually my insertion sort is actually slower when running in an app with optimizations.  Closing the PR.

PR #905, opened 2019-11-27
Fix incorrect time complexity

fterh:
Time complexity of remove() is O(n + log n), which simplifies to O(n).

### Checklist

### Description

Fixes incorrect time complexity.

### DIFF
--- a/Heap/Heap.swift
+++ b/Heap/Heap.swift
@@ -212,7 +212,7 @@ extension Heap where T: Equatable {
     return nodes.index(where: { $0 == node })
   }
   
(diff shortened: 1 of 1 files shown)


kelvinlauKL:
Hey, thanks for the suggestion!

In this case, the time complexity is actually O(log n). This is because `remove` does not need to look at all _n_ elements for it to complete the operation. 

I'm closing this now, but feel free to leave a comment if you need additional clarification.

kelvinlauKL:
Whoops, I was looking at the wrong `remove` operation. So you're right! Thanks for the catch.

PR #907, opened 2019-12-10
Update simann_example.swift

TRGoCPftF:
### Checklist

### Description

 
Example file was not functional as found.

Corrected recursive definition where shuffle() was invoking self.shuffle() instead of shuffling the point array `self.tour`, which left the example in an infinite loop of shuffle() invocation.


richard-ash:
Thanks for the fix!

Issue #864, opened 2019-03-01
Keeping a Tail pointer & Count to LinkedList implementation


biscottigelato:
Any reason why a tail pointer and/or a count value is not kept in the Linked List implementation? Right now .last will actually traverse through the entire linked list at O(n) just to get at the last node. Appending to the linked list also traverses the entire linked list at O(n). Count will also cause a Linked List traversal at O(n)

Adding a tail pointer and a count value adds O(1) to space and performance at most for affected operations, however reduces .last, .append and .count from O(n) to O(1).

Happy to do a pull request on this if people are interested

biscottigelato:
I am thinking more from the code/implementation perspective. Didn't really consider how this would impact the article. I can also add a blurb about adding tail and count and how that can improve efficiency. But that requires the pre-text of the tail/count code be part of the implementation tho, otherwise it won't make too much sense nor enough of a difference.

hollance:
But you can show how ro change the implementation to add those things. And then include this as a different Swift file, LinkedListWithTailPointer.swift.

kelvinlauKL:
Yup, as you suggested having a tail reference can improve time complexity. The article (and most introductory courses) don't talk about the tail until later on. 

It's better for the reader to understand what inefficiencies they have to deal with and why the tail reference is introduced in the first place.

PR #909, opened 2019-12-20
updated Linked List by optimize tail node and count

hupei1991:
### Checklist

- [x ] I've looked at the [contribution guidelines](https://github.com/raywenderlich/swift-algorithm-club/blob/master/.github/CONTRIBUTING.md).

### Description

 
I updated the Linked List data structure. Because the time complexity of previous version is O(n) to get the tail node and count the total number of node. I updated this to get these in time complexity of O(1). Because this is the actually optimal solution for double linked list, which make O(n) solution a little bit misleading.

### DIFF
--- a/Linked List/LinkedList.playground/Contents.swift
+++ b/Linked List/LinkedList.playground/Contents.swift
@@ -24,16 +24,17 @@ public final class LinkedList<T> {
     /// The head of the Linked List
     private(set) var head: Node?
     
-    /// Computed property to iterate through the linked list and return the last node in the list (if any)
+    /// Return the first node of the Linked List for public scope
+    public var first: Node? {
+        return head
+    }
+    
+    /// The tail of the Linked List
+    private(set) var tail: Node?
+    
+    /// Return the last node of the Linked List for public scope
     public var last: Node? {
-        guard var node = head else {
-            return nil
-        }
-        
-        while let next = node.next {
-            node = next
-        }
-        return node
+        return tail
     }
     
     /// Computed property to check if the linked list is empty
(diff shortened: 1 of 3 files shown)


kelvinlauKL:
Hey, thanks for the suggestion!

You're right to say that the current implementation is not the optimal solution for a LinkedList. However, the author intentionally left out the tail reference for the sake of showing what operations need to happen without it. 

If you really want to include the tail reference, we'll need to do a bigger refactor of the text to help the prose flow from the tail-less world.

hupei1991:
Should we do the refactoration upon the tutorial text? Because when someone was trying to use this piece of code in their own code base, the original implementation would be misleading and even causing performance problem.

kelvinlauKL:
To be fair, I don't think the current tutorial is misleading. The goal is to introduce the reader to the singly linked list - which includes all of the challenges with backwards iteration. The author notes these drawbacks in the beginning:

>A singly linked list uses a little less memory than a doubly linked list because it doesn't need to store all those previous pointers.
>
>But if you have a node and you need to find its previous node, you're screwed. You have to start at the head of the list and iterate through the entire list until you get to the right node.
>
>For many tasks, a doubly linked list makes things easier.

I'd like to preserve the author's original intention of showing what a singly linked list is about. There's also a few notes throughout the text alluding to how a previous / tail reference could optimize the performance of the list. IMO this approach does provide value to the reader.

Perhaps we could just add a new file with the optimizations and add a section near the end telling the reader what the optimized version looks like?

If you're up for a bigger task, adding a separate entry for the doubly linked list might make sense as well.

kelvinlauKL:
#864 also discusses this

richard-ash:
Closing due to inactivity. If you reopen we're happy to take another look :)

PR #912, opened 2020-01-22
Typos corrected

gmoraleda:
### Checklist

### Description

Minor typos corrected.

### DIFF
--- a/Count Occurrences/README.markdown
+++ b/Count Occurrences/README.markdown
@@ -22,13 +22,13 @@ The trick is to use two binary searches, one to find where the `3`s start (the l
 In code this looks as follows:
 
 ```swift
-func countOccurrences<T: Comparable>(of key: T, in array: [T]) -> Int {
+func countOccurrences<T: Comparable>(of key: T, in a: [T]) -> Int {
   var leftBoundary: Int {
     var low = 0
-    var high = array.count
+    var high = a.count
     while low < high {
       let midIndex = low + (high - low)/2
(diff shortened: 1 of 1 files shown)


gmoraleda:
> I think your edit made the code more ambiguous

It's consistent with the rest of examples and the function content. The parameter name (`array`) and the signature didn't correspond to the function content nor the way it was being called.

Count Occurrences/README.markdown
```diff
@@ -22,7 +22,7 @@ The trick is to use two binary searches, one to find where the `3`s start (the l
 In code this looks as follows:
 
 ```swift
-func countOccurrences<T: Comparable>(of key: T, in array: [T]) -> Int {
+func countOccurrences<T: Comparable>(of key: T, in a: [T]) -> Int {
```
kelvinlauKL: `array` would be preferable here. I would refer to some of the principles in the Swift API design [guidelines](https://swift.org/documentation/api-design-guidelines/) for this:

> Clarity is more important than brevity. Although Swift code can be compact, it is a non-goal to enable the smallest possible code with the fewest characters. Brevity in Swift code, where it occurs, is a side-effect of the strong type system and features that naturally reduce boilerplate.

kelvinlauKL: The README is not accurate with the actual code. All the `a` references should actually be `array`.

gmoraleda: Agree. Array is clearer. Just exchanged the renaming.

PR #918, opened 2020-03-06
Modified Trie to add prefix support to boolean contains(), plus unit …

gentlesystems:
…tests.

### Checklist

### Description

 
There was already a function for a rapid search to see if the trie contains a word, but there was not an equivalent function that returned in the same (O) time to check for prefix matches.  The existing function findWordsWithPrefix() takes significantly longer because it tracks down all terminating nodes in the subtrie, and is a waste if you're just looking to know if the prefix matches.  A classic use case of this type of boolean request is to determine whether to return a document that matches a find-as-you-type prefix search on a pre-constructed trie of its tokenized contents.

I modified the contains() function to support this functionality for a project a couple years ago, and as I was updating that project it occurred to me that I ought to contribute this minor change upstream.  I've built similar classes in other languages, but your swift version was so clear and elegant that I just used it instead of porting another or rewriting it.

Unit tests were simply adapted from the existing findWordsWithPrefix() tests.

### DIFF
--- a/Trie/Trie/Trie/Trie.swift
+++ b/Trie/Trie/Trie/Trie.swift
@@ -117,9 +117,12 @@ extension Trie {
 
   /// Determines whether a word is in the trie.
   ///
-  /// - Parameter word: the word to check for
+  /// - Parameters:
+  ///   - word: the word to check for
+  ///   - matchPrefix: whether the search word should match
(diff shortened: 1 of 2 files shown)


richard-ash:
Thanks for the merge!

PR #922, opened 2020-03-27
updated depricated hashable values to hash into methods

mick31:
### Checklist

- [x ] This pull request is complete and ready for review.

### Description
This PR resolves the `'Hashable.hashValue' is deprecated as a protocol requirement` by updating both the `Vertex` and `Edge` structures to  `implementing 'hash(into:)' instead`.

There are many other instances of this warning that need to be addressed but I believe those should be updated on a case by case basis to help keep PR's as small and concise as possible 😁 

##### Vetex
image

##### Edge
image

### DIFF
--- a/Graph/Graph/Edge.swift
+++ b/Graph/Graph/Edge.swift
@@ -29,12 +29,10 @@ extension Edge: CustomStringConvertible {
 
 extension Edge: Hashable {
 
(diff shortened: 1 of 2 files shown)


richard-ash:
@mick31 thanks for the update! Looks good to me 🙌

PR #934, opened 2020-05-31
Updating Counting Sort for stability

JMatharu:
### Checklist

### Description

 
On step 3 of Counting sort we need to Loop through the array in reverse to keep the stability of the new array
(i.e. 7, is at index 3 and 6, thus in sortedArray the position of 7 at index 3 should be before 7 at index 6

Also updated Tests project to support Swift 5


richard-ash:
Thanks for the update to the algorithm @JMatharu!

PR #949, opened 2020-10-01
Implemented binary indexed tree.

st235:
Implemented Binary Indexed (Fenwick) tree. Fenwick Tree is a data structure that can efficiently update elements and calculate prefix sums in a table of numbers.


st235:
Hi Algorithm Club 👋

Thank you for such a nice and unique repository! I really appreciate what you do here and really love algorithms like you. I found there are no Fenwick tree data structure and implemented it on my own to make this repo a little bit better. Hopefully, you will like the proposed changes.

PR #969, opened 2021-03-05
Auronas: remove force casts on dequeuing

Shahn-Auronas:
### Checklist

### DIFF
--- a/A-Star/AStar.swift
+++ b/A-Star/AStar.swift
@@ -54,10 +54,7 @@ public final class AStar<G: Graph> {
     /// - Precondition: both `source` and `target` belong to `graph`.
     public func path(start: G.Vertex, target: G.Vertex) -> [G.Vertex] {
         open.insert(Node<G.Vertex>(vertex: start, cost: 0, estimate: heuristic(start, target)))
-        while !open.isEmpty {
-            guard let node = open.remove() else {
-                break
-            }
+        while !open.isEmpty, let node = open.remove() {
             costs[node.vertex] = node.cost
 
             if (node.vertex == target) {
--- a/Huffman Coding/Huffman.swift
+++ b/Huffman Coding/Huffman.swift
@@ -95,8 +95,7 @@ extension Huffman {
         while queue.count > 1 {
             // Find the two nodes with the smallest frequencies that do not have
             // a parent node yet.
-            let node1 = queue.dequeue()!
-            let node2 = queue.dequeue()!
+            guard let node1 = queue.dequeue(), let node2 = queue.dequeue() else { return }
 
             // Create a new intermediate node.
             var parentNode = Node()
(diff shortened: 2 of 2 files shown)


hollance:
It's not really necessary to do this, since the `!` is guaranteed to work.

Huffman Coding/Huffman.swift
```diff
@@ -95,8 +95,7 @@ extension Huffman {
         while queue.count > 1 {
             // Find the two nodes with the smallest frequencies that do not have
             // a parent node yet.
-            let node1 = queue.dequeue()!
-            let node2 = queue.dequeue()!
+            guard let node1 = queue.dequeue(), let node2 = queue.dequeue() else { return }
```
snooky23, line 98: for better readability, please split node1 and node2 into two lines.

A-Star/AStar.swift
```diff
@@ -54,10 +54,7 @@ public final class AStar<G: Graph> {
     /// - Precondition: both `source` and `target` belong to `graph`.
     public func path(start: G.Vertex, target: G.Vertex) -> [G.Vertex] {
         open.insert(Node<G.Vertex>(vertex: start, cost: 0, estimate: heuristic(start, target)))
-        while !open.isEmpty {
-            guard let node = open.remove() else {
-                break
-            }
+        while !open.isEmpty, let node = open.remove() {
```
snooky23, line 57: for better readability, please split `!open.isEmpty` and `let node = open.remove()` into two lines.

snooky23 (changes requested):

Huffman Coding/Huffman.swift
hollance: Again, no need to do this, plus in my opinion it doesn't actually improve readability.

hollance: Force unwrap is perfectly acceptable when it cannot possibly fail. Writing code that handles errors that can never occur is more dangerous because it can hide subtle bugs or cause people to insert subtle bugs later (i.e. it adds complexity that doesn't need to be there).

Shahn-Auronas: Not sure I follow the complexity argument in using guard statements. I take your point on force unwrap when it's impossible to fail, ergo my point above on tests. My point was good coding style - i.e., not getting in the habit of force unwrapping when you can easily guard. @snooky23 matter of taste, but I don't disagree with you.

hollance: When the guard fails, you return from the method. This can never happen, but you've added logic that makes it seem like returning early from the method is allowed. But it isn't, since there is logic that happens behind the while loop. Your change doesn't introduce a bug by itself because the return will never happen, but it makes it *appear* as if it's OK to exit from the method inside the while loop, which is not OK. Hence, you've added complexity that wasn't there before, and the potential for bugs to be introduced because the error handling path of your code is incorrect. It would be better to do a fatalError instead of a return, but that is also more complex than just force unwrapping.

A-Star/AStar.swift
hollance: I actually like this change. 😄 Having it all on one line would be my preference.

PR #979, opened 2021-06-30
fixes label-less method parameter

MartinP7r:
The following code samples all use `append(_:)` without the `value:` label and since Swift 3 (https://github.com/apple/swift-evolution/blob/master/proposals/0046-first-label.md) this has to be implemented with a preceding `_` in the function signature.

### Checklist

### Description

This is an update to one of the source code samples.  
Your README.md states compatibility with Swift4.2, so this change from Swift 3 is warranted, I hope.
Feel free to deny the PR if I should have overlooked something.


iiicey:
邮件已收到。我会尽快回复！祝您一切顺利！兰雨田

